PR #73, opened 2026-09-13
engine: kill a job's process group with killpg, never through the kill command

unarbos:
## Critical (QA `qa-020`)

`kill_job` (`crates/arbos-engine/src/tools/bash.rs:448`) shelled out to `kill -9 -<pid>` without `--`. procps `kill` reads the negative pid as an option and sends `kill(-1, SIGKILL)`: **every process the user may signal**.

On ArbosLife this happened twice in one night (00:24 and 03:09 UTC, 2026-09-13) from the QA loop's own kernel: the `bench-screenshot` scenario's background `python3 -m http.server 8765` hit `timeout_ms`, the kernel killed the job, and pm2, the Affine validator, the user manager and every SSH session of `const` died with it (validator down 00:24 → 04:29). The Affine agent traced it with `strace`, wrote up the evidence in `~/arbos-qa/deploy/DISABLED-BY-AFFINE-AGENT-2026-09-13.md` on the box, and disabled the loop. Any `arbos-kernel` whose background job times out has the same effect on any machine with procps `kill`.

## Fix

No shell any more:

- `libc::killpg(pid, SIGKILL)` on the job's own process group (jobs are spawned with `process_group(0)`, so the leader pid is the group id), `libc::kill(pid, SIGKILL)` as the fallback.
- pid `0`, `1`, or anything that does not fit `pid_t` is refused outright (`kill(0)`/`killpg(0)` mean "my own group", 1 is init).

The two messages that tell the model how to stop a job by hand already say `kill -- -{pid}` and are unchanged.

## Tests (`bash.rs`, `kill_tests`)

- A job group (`sh -c 'sleep 300 & wait'` in its own process group) is killed while a sleeping bystander in another group survives, the exact thing `kill -1` would have taken.
- `kill_job(0)` and `kill_job(1)` are no-ops (if either were signalled the test process would not survive to assert).

## Ask

This should go into `rust` and into the integration branch #58 before either runs anywhere shared. The QA loop on ArbosLife stays off until it does, and should then run as its own Unix user (see the note on the box).

<div><a href="https://cursor.com/agents/bc-f2e2f30d-1298-59f1-a24c-55113322de28?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f2e2f30d-1298-59f1-a24c-55113322de28&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: superseded. The same fix landed as #74 (`a840e28`, killpg on the job's process group). Content is in `main` (`7ea2bb3`, tag `v0.2.0` at `ab7cf62`); this branch conflicts with `rust` and carries nothing further. Closing.

PR #279, opened 2026-09-16
desktop: tool-call markup written as prose never shows — cut from the live stream and the settled line; a markup-only reply leaves no bubble

unarbos:
Companion to #278 (the kernel cuts the markup from the settled `assistant` line). Cursor never shows its own tool markup as prose; neither does the window now, at any moment of a reply.

## What was true before

- The live `assistant_delta` stream carried the raw `<invoke …>` tokens into the reply item as they arrived, and they stayed on screen until the settled line replaced the item through #252's step pairing.
- When the kernel's settled line was **empty** (a reply that was only markup), the pairing kept the streamed text — `if !text.is_empty() { *body = text }` — so the tags stayed for good, in an otherwise empty turn.

## Change

- `src/markup.rs`: `strip_live(text)` cuts the same families the kernel knows — `<function_calls>`/`<invoke …>`, `<tool_call>`/`<tool_calls>`, `<function_call>`, `<function=name>`, `<|python_tag|>`, `<|tool_call|>`, `[TOOL_CALLS]` — from the opener to its closer, or to the end while the closer has not arrived; an opener still arriving at the very end (`<inv`) is held back so no tag flickers in. Prose that mentions `invoke`, HTML, and `a < b` are untouched (unit tests).
- The session keeps each live reply's raw text in `stream_raw` (so delta merging keeps working) and shows the cut version; both the `step`-paired delta path and the older chunk path go through it.
- The settled line replaces the item as before; a **settled line that is empty drops the item** (indices in the step maps and `streaming_agent` shift with it), so a markup-only reply leaves no bubble. The settled line is also run through the cut, which covers a kernel from before #278.

## Verified live (Linux, `main` kernel, i.e. before #278)

Two prompts on a cold place:
- "Reply with exactly this text … `<function_calls><invoke …>…`": the kernel's settled line was the raw markup; the window shows **no bubble** for that turn — no tags, no empty card.
- "Write 'Listing the folder now.' and then, as literal text, `<invoke …>`": the window shows **"Listing the folder now."** only; the kernel's nudge then made the model call `bash`, and the answer followed. No frame at t2/t8/end shows a tag.

## Gate (done)

`ui_pass.py` on this build (`media/qa-ui/live-markup-2/`): **110 pass**, 2 standing fails (`composer-context`, `work` fold toggle) and `rewind-turn`, which I chased: on both this build and the cycle-14 build the kernel answers "no checkpoints yet" when the earlier turns did not touch files — a model-behaviour flake since the gate moved to Gemini (openai/* through the rig's OpenRouter key is refused with a 403 policy block as of tonight), not this change. `weight-visible` passes. `cargo check --locked --all-targets` clean; `cargo test --lib` 55 pass (5 new in `markup::tests`); CI green on all three jobs.


unarbos:
Merge steward: holding this one for now — it is green and the diff is fine (desktop-only, companion to #278 which is now on `main`), but the body says the gate run is still pending ("Gate run noted below when it finishes"). Add the gate result and mark it ready, and it goes in on the next pass.

PR #302, opened 2026-09-16
ios: cycle 9 — the phone's push half: permission at the first message, the token to the hub on every attach, an honest push state, a tapped push lands on its project

unarbos:
## What

The app's half of #301.

- **Permission at a sensible moment.** iOS asks when the user sends their first message in a project — the moment "tell me when it answers" makes sense — not on the first frame. A phone that already said yes re-registers on launch without a prompt (tokens rotate).
- **The token goes to the hub** as `{"type":"push","platform":"apns","token":…,"sandbox":…}` on every live attach (sandbox for debug builds), and again when it changes. The hub's `pushed {enabled}` is kept.
- **Honest state in Settings** ("Notifications"): *On — the hub pushes replies, questions and failures to this phone* / *Banners only while the app is open — the hub has no Apple push key yet* / *Banners only while the app is open: this build has no push entitlement* / *Off — allow notifications for Arbos in iOS Settings* / *Off until you send your first message — then iOS asks*. Never pretends a push will arrive.
- **A tapped push opens its project** — the same `userInfo.target` path as local banners, including from a cold start.
- **CI signs with `aps-environment` the day the App ID has the Push Notifications capability**: a step asks App Store Connect for the bundle ID's capabilities and passes `CODE_SIGN_ENTITLEMENTS=Arbos/Arbos.entitlements` only then, so an export never fails on a profile without it, and no release change is needed when Jacob's key and tick land.
- A live `notify` with the chat in front is seen at once (and every client hears so) instead of joining the away card. A hub from before #301 that lets `push` reach the kernel produces a silent `enabled: false`, not a red line.

## How to check

- [x] Simulator build on the loop's EC2 Mac (Xcode 27.0, iOS 27.0): **BUILD SUCCEEDED**
- [x] Fresh install: open a project → no prompt; first send → the system prompt; Allow → Settings reads "Banners only while the app is open: this build has no push entitlement" (a simulator debug build). Stills `media/mobile/cycle-9/01`, `02`.
- [x] `xcrun simctl push` with the hub's exact APNs payload from #301's note: banner and badge 2 on the home screen; banner over another app; tap → `longproj`'s chat. Stills `03`–`05`.
- [x] A live reply with the chat in front: bubble only, no away card (`06`).
- [ ] A real `pushed {enabled}` from a #301 hub and a real token: needs a TestFlight build with the entitlement on a phone — after Jacob's tick. Until then Settings shows the entitlement line, which is the truth.

## Notes for the reviewer

- The entitlements file is `ios/Arbos/Arbos.entitlements` (`aps-environment = production`); it is not referenced by the project — CI passes it as a build setting when the capability exists. Simulator builds never get it.
- Ledger rows M-74 … M-76, and M-70 closed, in the project store's `internal/mobile-findings.md`.

<div><a href="https://cursor.com/agents/bc-08d8261b-fea2-5075-9949-d45f6f9d4acc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-08d8261b-fea2-5075-9949-d45f6f9d4acc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: merged, but the first TestFlight run on top of it (`b76ea1d`, run 35089444444) failed in the new "Push capability on the App ID?" step — one second, no output. `pip3 install … >/dev/null 2>&1` fails under `bash -e` on the macOS runner (pip refuses a system install), so the shell exited before Python ran. #305 makes the step fail-safe (a failed install or lookup leaves `ENT` empty and the build ships without the entitlement, as intended). If you prefer a different way to get pyjwt/requests onto the runner, follow up on top.

PR #308, opened 2026-09-16
update: a kernel payload in the feed, and a command that installs it

unarbos:
## What

The feed can describe a kernel, CI publishes one, and `arbos-updatectl kernel` installs it — verified, with rollback, with a person choosing the moment.

Slices 2 and 3 of `docs/kernel-self-update-design.md` in the agent store. Nothing runs unattended yet, and nothing in the kernel changes.

## Why

*"Shouldn't we auto update the kernels when we connect."* Jacob's `arboslife/subnet120` is running days-old code and answering "hello" with a directory listing. It is hub-attached, so nobody ssh's into it and the #229 install-on-connect path never runs; nobody attaches to it either. Both existing ways of updating a kernel miss it.

**This PR fixes that machine today**, by hand, with the full verification chain — before anything is automatic.

## The feed can describe a kernel

`Download` gains a `component`:

```json
{ "platform": "linux", "arch": "x86_64", "component": "kernel",
  "url": "…/arbos-kernel-0.2.0-903-linux-x86_64.tar.gz", … }
```

It defaults to `app` and is **skipped when it is `app`**, so every feed published so far is byte-for-byte what it was, and an app that predates the field reads a newer feed unchanged. There is a test for exactly that, because it is the property that lets this ship without a flag day.

CI publishes `arbos-kernel-<version>-<build>-linux-x86_64.tar.gz` beside the app tarball — **the same binary from the same build**, a second archive rather than a second compile, so the two can never be from different code. A headless box has no use for the 32 MB of which the desktop is most.

## The mechanism

`arbos_update::kernel` is the whole thing with **the moment left out**, deliberately. A kernel is serving live agents and the only safe instant is between turns with nothing waiting on a human — which the kernel's own `idle::verdict` already answers and this crate cannot see. Every function here is "is this allowed, and do it"; the caller owns when.

**The commit decides, not the version.** `0.2.0` has not moved in weeks, so semver alone cannot tell this morning's kernel from last week's — which is precisely how one went stale without anyone noticing. So the rule is the one the ssh path has used since #229: older version, or the same version from a different commit, is behind; a newer one is left alone.

The install is the same two renames the app uses, so the same promise holds: **there is no moment at which the binary is half replaced**. The new binary is run before it is moved, and run again once it is in place; a failure at either point puts the old one back through `Swap`'s `Drop`.

## Saying no

Four refusals, and they are *answers*, not errors — a caller reports them rather than retrying:

| refusal | why |
| --- | --- |
| a binary under `target/debug` or `target/release` | it is somebody's working copy; replacing it would swap out the thing they just built and be undone by their next build |
| `--pin` | held deliberately |
| a newer build than the channel's | almost always a local build; the ssh path's existing rule |
| no kernel for this platform | the channel simply does not publish one |

## Checked against the real thing

Unit tests cover the parsing, the comparison and the refusals. The parts that matter were also exercised against a real 250 MB `arbos-kernel` binary and the live dev feed:

```
running   0.1.40 oldsha012345
verdict   behind, so it will be replaced
after     0.2.0 181b657fc404
beside it now: arbos-kernel
```

- **The same payload, once the commit matches** → `already the newest build on this channel`.
- **A payload that will not execute** → the running kernel is byte-identical afterwards. This found a real leak: the staging directory survived the failure, so a second attempt would trip over the remains. Fixed, and the cleanup now runs on every path.
- **Against the live `dev` feed**, a binary in this checkout → `is inside a build tree, so it is somebody's working copy and will not be replaced`. A copy outside one → `the channel publishes no kernel for this platform`, which is correct until this merges and the next publish carries one.

`cargo test -p arbos-update`: 58 passed. `cargo build --workspace --locked` and `cd desktop && cargo check` both clean.

## How to use it on `subnet120`

Once this is on `main` and one publish has run:

```
arbos-updatectl kernel --channel dev --binary ~/.cargo/bin/arbos-kernel
arbos-updatectl kernel --channel dev --binary ~/.cargo/bin/arbos-kernel --install
```

The first reports; the second replaces. The running process keeps the old image until it restarts — the command says so rather than pretending otherwise, and restarting is the supervisor's job (`kill -TERM` the pid in `<place>/.arbos/runtime/kernel.json`; the tmux loops restart it).

## Notes for the reviewer

- **`reqwest` is a new direct dependency** of `arbos-update`, for fetching the feed and the payload in the tool. It is the same client and TLS backend `arbos-kernel` already uses and it was already compiled in both trees, so it is a manifest line rather than a second HTTP stack.
- **`Feed::newest`** is new beside `available`. The app compares on the version pair and refuses anything not strictly newer (downgrade protection); a kernel has no build number to compare with, so it asks for the newest and decides by commit. Two callers, two correct rules, one feed.
- **The `--artifact` spec grew an optional third part**, `<platform>/<arch>[/<component>]=<path>`. Every existing invocation means what it meant.
- **`Args` learned switches.** `--install` takes no value; the parser now treats a flag whose next word is another flag, or which ends the line, as a switch. Tested both ways round.
- Restart semantics are still with the features agent (`internal/features-inbox/2026-09-16-kernel-self-update-restart-semantics.md`). Nothing here restarts anything, which is why it can land first.

## Next

The design's slice 1 — making skew **visible where Jacob already looks** rather than only in a log. The local half needs no kernel change at all: `kernel.json` already carries `version` and `git_sha`, so the desktop can compare them against the feed's newest commit and say so in the bar. The hub roster half needs the two fields the features agent is being asked about.


unarbos:
Merge steward: `main` is now `1e09d1b` — #308 merged; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `1e09d1b` (arbos-update crate + workflows; desktop Cargo.lock moved).

PR #307, opened 2026-09-16
Version skew visible: git_sha and built_at on hello, /healthz, Register; idle::update_verdict for the self-updater

unarbos:
## What

`features-inbox/2026-09-16-kernel-self-update-restart-semantics.md` §3: semver has sat at `0.2.0` for weeks, so `hello` could not tell this morning's kernel from last week's — the skew a self-updater, or a person looking at `GET /list`, has to be able to see at all. Plus the restart gate the updater will call.

## Change

**Fields**
- `build.rs` bakes `ARBOS_BUILT_AT` (`YYYY-MM-DDTHH:MMZ`, UTC to the minute, no dependency) beside `ARBOS_GIT_SHA`; `klog::built_at()` reads it.
- `Frame::Hello { …, kernel, git_sha, built_at, … }` — both `#[serde(default)]`, omitted when unknown.
- `GET /` and `/healthz` on the attach port say the same.
- `HubFrame::Register { …, version, git_sha, built_at, … }`; the hub keeps the newest registrant's on `MachineEntry` and shows them on `MachineInfo`. The hub reports what each machine *is*; it does no ordering — staleness is the client's call, since only the client holds the feed. The hub relays raw JSON since #275, so an old hub drops nothing.
- Named `built_at`, not `build`: the update feed's `build` is a commit count, and a count compared with a timestamp yields a number rather than an error.

**Gate** — `idle::update_verdict(hooks, horizon_ms)`: `verdict()` with three corrections for a kernel restarting itself. A parked ask is not a reason to wait (the question is a file, the answer arrives as an inbox file, the new kernel opens the turn) so `Waiting` reads as `Idle`; a pending approval already reads as `Busy` (its turn is blocked inside the tool), so `clear_approves` at the next boot never sees one that mattered; a running detached job is `Busy` (the job's leash kills it the moment its kernel dies; `keep` only spares the boot reap) and so is a remote child mid-turn. `--until-idle` is unchanged.

## Test

`hello_built_at_e2e`: `hello` carries a hex sha and a `YYYY-MM-DDTHH:MMZ` `built_at`; `/healthz` agrees. `idle::update_verdict_tests`: an ask lets the update through and still holds `--until-idle`; a turn and a running job hold the gate. Core round-trip, hub, port_health, hello_identity, leash suites green; desktop compiles.

[...]

PR #315, opened 2026-09-16
desktop: the delegated-worker turn drawn once (F-82), the keyless first line, the reconnect double, status lines in every form, the front tab restored, and the acceptance journey harness

unarbos:
## F-82 — Jacob's "write and run bubble sort"

His screen drew one worker three times while the root waited on it: a live `Working Starting worker… ⌄` header whose chevron did nothing, a Working / Stop All card over the composer, and the `Working 1` pill — with a paragraph reading `status: Delegating to worker…` between them and a long empty gap above the card.

Cursor, run live today on the same prompt (Project chat, one delegated worker): the prompt, its prose, **one line** `1 Working  Delegating bubble sort task`, and the `Working 1` pill. With three workers: one line `3 Working  Waiting on three writers`. **No card** — it sits behind the pill and never opens by itself (F-13's card from the cycle-7 pair was the pill's disclosure, not the default).

Now:
- while the root's turn waits on workers, one `N Working  <step>` line (the one worker's live step, else the coordinator's, else its title), shimmering; a click opens the one worker, or the Working card for several. Per-worker lines return once the root's turn has ended and workers run on in the background (Jacob's earlier Cursor stills). A worker just spawned counts as working from the spawn, as the pill does (F-86).
- worker lines sit after the turn's prose, the footer last, where Cursor draws them (F-85).
- the live `Working <step> ⌄` header draws only over rows it can fold (a thought, a run with tool rows, a checklist), and its chevron folds them. Over prose alone or a wait on workers there is no header; the step is the heartbeat's `Working <step>` line (Cursor's form) or the workers' line.
- the Working card opens on the pill (or the line) and closes on ×. `Workspace.working_card_open` replaces the root view's closed-set.

Pairs (Cursor left, Arbos right), in the project store: `media/cursor-reference/cycle-17/f82-pair-before-cursor-left-arbos-right.png`, `f82-pair-after-cursor-left-arbos-right.png`; Cursor's three-worker frame `f82-cursor-three-workers-one-line-no-card.png`.

## Status lines in every form (F-92, QA `qal-j01`)

[...]

PR #318, opened 2026-09-16
kernel: arbos-kernel update, a verified one-command update

unarbos:
## What

`arbos-kernel update` — the whole verified update, with a person choosing the moment, and an honest answer about what is actually serving.

Slice 3 of the kernel self-update design. **This is the command that fixes `subnet120` today.**

## Why a command before anything automatic

`arboslife/subnet120` is running days-old code. Nobody ssh's into it, so the #229 install-on-connect path never runs; nobody attaches to it either. Both existing ways of updating a kernel miss exactly that machine.

`arbos-updatectl kernel` ([#308](https://github.com/unarbos/arbos/pull/308)) could already do this, but `arbos-updatectl` is not on that box — the kernel is.

## Replacing the binary is not updating the kernel

This is the part that grew after a look at the live machines, and it is the more important half.

`subnet120`'s binary had **already been replaced on 15 September** while the running process kept the old image — `/proc/<pid>/exe` reading `(deleted)`. That is how it stayed stale with nobody noticing. A command that answered "up to date" from the file alone would have reported that machine as current while it served old code.

So it reports both:

```
running   0.2.0 9d1fca4bf43b (/tmp/.../bin/arbos-kernel)
channel   dev
available 0.2.0 (1021) (2cf7d5b) published 2026-09-16T13:29:19Z
          Merge #311: desktop/qa/parity — the symmetry loop's rig tooling lives in the repository
          https://github.com/unarbos/arbos/releases/download/dev/arbos-kernel-0.2.0-1021-linux-x86_64.tar.gz

serving   0.2.0 181b657fc404 (pid 171347) in /tmp/.../place  ← older than the binary on disk
```

`--place` names places to look at and defaults to the working directory when it is one. Each running kernel is reported **as it described itself in `kernel.json`** — the version and commit that process compiled in, which is the only honest answer to what code is running there.

I reproduced the `subnet120` state to check it: a current binary, a live pid and a `kernel.json` from an older build gets the marker; matching commits do not; a dead pid is not serving anything.

## Run against the live feed, not just tests

Full chain — live `dev` feed, the real signed payload, the public key committed in #263:

```
running   0.1.40 aaaaaaaaaaaa
available 0.2.0 (1021) (2cf7d5b)
fetching  8673555 bytes
installed 0.2.0 2cf7d5b9b496
```

The binary on disk then reports `arbos-kernel 0.2.0 2cf7d5b9b496 protocol 1`, with nothing left beside it. A second run says `already the newest build on this channel`; `--pin 0.2.0+879` says `pinned to 0.2.0+879`.

**8.6 MB, not 32** — #308's kernel-only artifact doing its job.

## What it refuses

| refusal | exit |
| --- | --- |
| the binary is under `target/debug` or `target/release` — somebody's working copy | 0 |
| `--pin` holds it | 0 |
| the running build is newer than the channel's | 0 |
| the channel publishes no kernel for this platform | **1** |

A decision is not a failure; only the last is worth a script noticing.

## The channel defaults by what the machine is

`dev` when the machine is registered with a hub, `stable` when it is not. A machine on the hub is part of the mesh and is meant to track `main`. `--channel` overrides it.

## It does not restart anything, and says so

```
The binary is replaced. Any kernel already running keeps the old code until it
restarts. Under a supervisor loop, stopping it is enough:

    kill -TERM $(sed -n 's/.*"pid":[[:space:]]*\([0-9]*\).*/\1/p' \
        <place>/.arbos/runtime/kernel.json)
```

Choosing that moment unattended is slice 4, with `idle::update_verdict` (#307) as the gate.

**One design change that came out of the same finding, for slice 4:** `subnet120`'s parent is `init`, so **nothing supervises it**. A kernel that exits there is gone. The unsupervised path is therefore `execv` rather than spawn-and-exit — same pid, no lock contention, and a failed `execv` leaves the old image serving instead of leaving nothing. That also removes the 30-second place-lock retry the features agent asked for, which was a consequence of spawning. And the path must be captured *before* the swap and exec'd by name: after a replacement, `/proc/self/exe` reads `(deleted)`, which is literally what was observed on that box.

## For the mesh worker, on `subnet120`

```
arbos-kernel update --channel dev --binary ~/.cargo/bin/arbos-kernel --place <place>
arbos-kernel update --channel dev --binary ~/.cargo/bin/arbos-kernel --place <place> --install
```

The first reports, the second replaces. Then restart it by hand with the same command and environment — it has no supervisor, so nothing else will.

## How to check

- [x] `cargo test --workspace --locked` — 663 passed, 0 failed.
- [x] End to end against the live `dev` feed and the committed key.
- [x] The `subnet120` state reproduced: stale-process marker, matching-commit case, dead pid.
- [x] Idempotent, and the pin holds.

## Notes for the reviewer

- **`net` moved into `arbos-update`** so the two programs that update a kernel fetch the same way. It deliberately does not decide whether a URL may be fetched — that stays `Download::check_url`, which every caller runs first.
- **`arbos-kernel` gains `arbos-update`** as a path dependency. `reqwest` and `serde` were already in its tree, so the new code is `ring` for the signature and nothing else.
- **`Args::parse` takes an iterator**, matching `rewind` and `worker`.
- `--pin` is per-invocation. The config form (`[update] channel`, `pin`) belongs with slice 4, where something reads it with no person present.


unarbos:
Merge steward: `main` is now `5219432` — #316, #314, #317, #315 and #318 merged; `rust` fast-forwarded to match. #318's first kernel job failed on `standing_pass_e2e::rewound_arrives_before_the_file_restore` (the known flake, untouched by this PR) and passed on rerun. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `5219432` — this head has `arbos-kernel update`.

PR #319, opened 2026-09-16
ios cycle 14 — Jacob's build-1021 feedback: photos drawn on the card, a dictated note waits for his send (and is kept out of the gateway's kernel), remote workers say where they run; M-88

unarbos:
## What

Four of Jacob's five TestFlight reports on build 1021 (`media/mobile/feedback/2026-09-16-17…21` in the project store). The fifth — the edge swipe back — is #322, because it cannot be verified in the simulator and the rest must not wait on it.

- **"The image sent does not show as an image in the chat section"** (16-19). The photo is drawn on the user card — a rounded thumbnail — before and after the kernel's echo and after a reopen. Bytes are cached under Caches/attachments by stored name (`AttachmentCache`); the kernel's `user` frame now decodes `attachments`, so a replayed line finds its picture. Files keep the paperclip line.
- **"The voice note was sent without me clicking send and then still shows in the chat section"** (16-21). Two causes. (a) #309 sent a dictated note by itself on the second mic tap; reverted — the words wait in the field for his send. (b) The voice gateway routes dictation finals to its own kernel as a turn (PR #56's answerer routing runs in dictation mode): dictating in `demo` without sending produced a turn in `pod`. Dictation now starts with `answerer: "model"`, which keeps the gateway off its kernel; verified `pod` stays quiet. Gateway ask filed.
- **"Creating sub agent does not run just says starting forever"** (16-18). Two halves. A worker on another machine sends no steps back; the line now reads "Running on arboslife · reports here when done". And (M-100, found by the journey) an older kernel's spawn record names the worker by its brief while every later frame carries the id the kernel made from it, so the phone kept two entries — the brief's, forever "Starting", and the id's, which went Done; the spawn record also arrives twice (call start, call end) and recreated the first. A brief-keyed worker is now adopted into the id the moment a frame names it, and the second record recognises it there. This is very likely the exact "starting forever" Jacob saw.
- **"Issue seeing text output"** (16-17). Empty assistant lines split one run of tool calls into three folds; they no longer break a run. Tool output itself is not in the kernel's record — ask filed.

From the loop:

- **M-88**: after a cold start with the hub unreachable, the list shrank to the single pod row and cached that. A hub that does not answer now keeps last time's rows marked Off and leaves the cache alone.
- **M-92**: those Off rows now come back by themselves — the roster retries at 10, 20, 40, then every 60 s until the hub answers.
- **M-97** (journey run 5): a line typed while a reply streamed closed the bubble and sat between the streamed words and the settled text — the reply's first word ("That") stood alone and the full reply landed after the steer as a second bubble; the card also rode above the reply it interrupted. Now a steer keeps the bubble open, the reply streams above the card, the card waits at the tail for its echo, and a settled reply reaches back past one pending card for the bubble whose words it begins with.
- **Push reason (#333)**: the `pushed` reply's `reason` is shown in Settings — "Banners only while the app is open — push is off: no APNs key on the hub…" in the hub's own words — instead of a fixed sentence. An older hub's silence still reads as the old-hub line.

## How to check

- [x] Simulator build on the loop's EC2 Mac (Xcode 27.0, iOS 27.0): **BUILD SUCCEEDED**
- [x] Photo card pending → echoed → reply, thumbnail throughout, and after a kill-and-reopen (`media/mobile/cycle-14/01-`, `02-`; journey run 6 `J6k-reopened.png`)
- [x] Dictation: words land in the field with the send arrow, nothing sent until the tap (`media/mobile/cycle-14/03-`)
- [x] Dictation in `demo` no longer puts a turn into `pod` (`04-`–`06-`)
- [x] M-100: one worker on `demo` (old kernel, `child: null`) — before: "Done" plus a "1 Working … · Starting" line and a "Working 1" pill that never cleared; after: "Done", no Working line, pill "Agents 14" (`media/mobile/cycle-18/01-`, `02-`)
- [x] Remote worker line reads "Running on arboslife · reports here when done" (code path: the spawn record's `machine` arg)
- [x] M-88: port 443 cut, cold start — four cached rows stay, marked Off (`media/mobile/cycle-15/01-`, `02-`)
- [x] M-92: same start; 45 s after the link returned all rows read Idle with no touch (`media/mobile/cycle-16/01-`)
- [x] M-97: steer typed 2 s into an eight-line reply — before: card above the reply, "Sending…" for the whole turn (`media/mobile/cycle-15/03-`); after: reply streams above, card at the tail, both replies whole (`04-`); journey runs 6–10 rendered steer + ask whole
- [x] Push reason: parsed from the `pushed` frame and rendered by `Notifier.pushLine`; the simulator has no push token so it shows the entitlement line there (`media/mobile/cycle-19/01-`) — the reason line is exercised on a device build against a hub with #333
- [x] Journey runs 5–10 on this branch (`internal/mobile-journey-runs.md`): every app-side step passes; the unverified ones are QA-rule cases

<div><a href="https://cursor.com/agents/bc-08d8261b-fea2-5075-9949-d45f6f9d4acc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-08d8261b-fea2-5075-9949-d45f6f9d4acc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: `main` is now `efcab58f` with these three commits (iOS only); `rust` fast-forwarded. Note for the record: the previous follow-ups (`6de1fad`/`410ed7c`) broke the iOS archive on `main` — `LiveKernelChat.swift:170` "missing condition in 'if'" — so the 03838b0 TestFlight run failed and shipped nothing. PR CI does not build iOS, only the TestFlight workflow on `main` does, so this could not have been caught before merge; `efc5c50` fixes it and I am watching the run for `efcab58f`.

PR #323, opened 2026-09-16
desktop: a prompt's echo is matched by the words awaited, not a clock (F-94 — prompts doubled after relaunch)

unarbos:
## F-94 — the journey's named bug

After quit and relaunch, prompts typed while the root was idle and its workers ran showed twice (J11 failed in 3 of 5 runs). The kernel's transcript held each once. The record after relaunch showed the pattern: each such prompt once with the kernel's clock and once without.

Cause: the window matches the kernel's `user` record to the card it drew by "the newest card with these words, within two minutes". The kernel can hold a typed line in its inbox for minutes — a turn still ending after Stop, workers running — and when it finally ran the line and wrote the record, the two-minute window had passed, so the record landed a second card. Live the second card sat at the end; after relaunch the local record's cards lose their clock and the history's stamped copies sort first, so the doubling read as `Continue(ts), Run(ts), Continue, Run`.

Now the session keeps the words of the prompts it sent and has not yet seen recorded (`awaiting_echo`); a `user` record with those words is the echo however long the kernel held it, and stamps the card with the kernel's clock. The clock rule stays as the fallback for records nobody awaited. First journey run on this build: J11 clean (`doubled_prompts=[]`); a second run is in progress and its line goes in a comment.

Stacked on #315 (both touch `foreign_prompt`'s neighbourhood); retarget after it merges.

## journey.py

J08 waits for the workers too — the coordinator may end its own turn at once and let them run on, and J09 then checked the disk before anything was written.

Ledger F-94 in `internal/symmetry-findings.md`.


unarbos:
Merged into `main` as 805851e1 (merge commit, right after #315). Closed rather than shown as merged because this PR's base was #315's branch; head `17d9114` is fully contained in `main`.

PR #326, opened 2026-09-16
JB-4: a stale window pin is loud once, over budget is said once per turn, a failed write is never "done"

unarbos:
## What

`internal/mobile-journey-runs.md` candidate JB-4 (the phone's `pod` kernel, 14:32 UTC). Cause, per the coordinator: a `window_tokens = 32000` pin in the pod's config.toml left over from a September 13 deploy, while the chat carried ~37k tokens — every step over budget with nothing old enough to compact, said thirteen times in one turn; a `write` failed and the reply still said "seeded". The pin is removed; three kernel behaviours around it still stood, and this fixes them.

## Change

**A stale pin is loud once, at start** (`serve.rs::warn_if_window_pinned_small`, `arbos_engine::standing_tokens`). When config.toml pins `window_tokens` below what a top-level agent's standing prompt needs (system prefix + tool schemas, doubled so a reply and a few turns fit), the agent's transcript gets one failed notice: `config.toml pins window_tokens = 32000, but this place's standing prompt is ~Nk tokens (system ~Nk, tools ~Nk) and needs a window of about Nk: every turn would run over budget with nothing to compact. Remove the pin (the model's own window is used) or raise it.` Plus a `window_pinned_small` log line. Once per pin value, not once per boot.

**Over budget is said once per turn** (`compact.rs`). The notice now names the window and the fix and is written and emitted only if the same words are not already on the transcript since the turn's wake.

**A failed write is never reported as done** (`turn.rs::unowned_failure`). When a step ends with a reply and the last tool batch had a failed write-class call (`write`, `edit`, `bash`, `apply_patch`, `git`, `pr`) with no later success of that tool, and the reply **claims a result** (`seeded`, `created`, `wrote`, `done`, `committed`, … or names the failed call's path) with no word of failure, one nudge: `Your last write call failed (<error>) and nothing since succeeded, but the reply reads as if it worked. Say what failed and what you will do, or fix it first — never report a failed write as done.` A reply that owns up, or that is about something else (a read-only helper whose side `write` the kernel refused, answering from its `grep`), is left alone — that second case was the CI failure in `helper_kinds_e2e`, fixed in the second commit.

## Test

`budget_once_e2e`: a 6000 pin → the failed notice at start with the sizes and "Remove the pin"; a four-step turn every step over budget → exactly one over-budget notice; a restart on the same pin → the pin notice still once, the second turn's over-budget once more. A write outside the place then "Seeded the project with tests/test_math.py" → the nudge, then the owned-up reply; one nudge. Full kernel suite green (97 suites), including `helper_kinds_e2e`.

## Not verifiable here

The run's transcript was not recorded in the store, so I could not see what the pod's `write` "exit 1" was; the nudge covers the visible symptom (a reply that claims what a failed write did not do) whatever the failure's cause.


unarbos:
Merge steward: holding — the kernel job failed twice (attempts 1 and 2) on `helper_kinds_e2e::an_explore_helper_runs_inline_read_only_and_its_answer_is_the_spawn_result` with `(replay: no more scripted replies)` as the helper's spawn result. That test is green on `main`, so this branch's `turn.rs`/`compact.rs` change looks to be consuming an extra scripted reply on the inline-helper path. Merges as soon as it is green.

PR #328, opened 2026-09-16
feedback: one turn's trajectory and log, redacted and bounded, for an in-app report

unarbos:
## What

Pre-empting the desktop's in-app feedback (its design, `docs/desktop-feedback-design.md`, has not landed yet; its owner's ask will follow). The kernel half it will need: hand over a turn's trajectory and the recent log in a form that is safe to send — redacted of credentials, bounded in size, covering the turn the user is complaining about rather than everything. Built here so every client sends the same thing.

Stacked on #327 (uses the tool record's `output` glance).

## The frames

Client → kernel:
```json
{"type":"feedback","agent":"root","seq":118,"note":"it printed my key"}
```
`seq` is any transcript line of the turn the user is looking at (omit for the latest turn); `note` is the user's words.

Kernel → client:
```json
{"type":"feedback_bundle","agent":"root",
 "turn":{"from":112,"to":119,"started_ms":…,"ended_ms":…,"complete":true,"lines":8,"of":8},
 "events":[…transcript lines, redacted and slimmed…],
 "log":[…kernel.log lines for the turn's span…],
 "kernel":{"version":"0.2.0","git_sha":"…","built_at":"…","os":"linux","arch":"x86_64","provider":"openrouter","model":"…","project":"…"},
 "note":"it printed my key",
 "redacted":{"secrets":0,"tokens":2,"values":0,"blocks":0},
 "truncated":false,"bytes":18234}
```

## What the kernel does

- **The turn, not everything:** from the wake holding `seq` to the next wake (or the end).
- **Slimmed:** tool `body` and `diff` out, the `output` glance in (filled from the body when an old record lacks it), `reasoning_details` out. Images and attachments stay as paths; no bytes.
- **The log for the span:** `kernel.log` lines with `ts` within the turn ±5 s, this agent's and the kernel's own (other agents' turns are not the complaint), last 400.
- **Redacted twice:** the kernel's own key and granted secrets by value (`arbos_engine::secrets::redact`, as transcripts already are), then credential *shapes* by form (`arbos_core::redact`, new): `sk-`/`sk-ant-` keys, `ghp_`/`github_pat_`, `xox?-`, `AKIA`/`ASIA`, `AIza`, `sk_live_`, `op://` refs, JWTs; `api_key = …`/`password: …`/`Authorization: …`-style values (whole-word key, so `window_tokens = 32000` stays); PEM private-key blocks, also when JSON-escaped on one line. Shapes not entropy, so a git sha stays. The `note` gets the same. `redacted` counts what went.
- **Bounded:** 256 KB; oldest log lines drop first, then the turn's middle (the wake, the user's line, and the last line always stay); `truncated` says so.
- Unknown agent → `error`. Answered on the requesting connection only, like `history`.

## Test

`feedback_bundle_e2e`: a turn whose `bash` prints an OpenRouter key and a GitHub token, then a second turn; `feedback` with a seq inside turn 1 returns only that turn, both keys gone (`[redacted:…]`), the tool body out and the glance in, the note redacted, the log for the span present, the build fields set; no seq → the latest turn; unknown agent → error. `redact::tests` (keys, tokens, values, PEM incl. one-line JSON; sha/URL/path/`tokens=12` kept), `feedback::tests` (turn span selection, slimming). Core, kernel unit, history/put suites green; desktop compiles (its catch-all lists the two frames).

## For the feedback owner

If the design wants a different shape — a file on disk instead of a frame, a different cap, the log as text — say so in the ask; the assembler is one function (`arbos_kernel::feedback::bundle`) and the redaction is reusable from the desktop (`arbos_core::redact::redact`).


unarbos:
Merge steward: `main` is now `e1837932` — #315, #323, #324, #325, #327, #328 and #319 merged, in that order; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `e1837932`.

PR #332, opened 2026-09-16
feedback bundle, revised: the exchange is the unit, budgets by outcome, fat args glanced, children and a tail

unarbos:
## What

The feedback owner's review of #328 (`internal/features-inbox/2026-09-16-feedback-bundle-desktop-answer.md`), all of it, with the later correction (`tail: N` instead of `turns: N`). #328 merged before this landed, so this is its follow-up on `main`.

## The three faults, fixed

**The span broke on any wake.** Spawn-first is the coordinator's normal shape — Jacob asks, root spawns, the turn ends, the workers report, a `done` wake opens the turn that answers — and a boundary on every wake put his words in one half and the answer in the other. Now the unit is an *exchange*: a `user`/`kickoff` wake to the next such wake; `done`, `job`, `serve`, `plan`, `say` wakes stay inside. With no `seq` the anchor is the last exchange the user opened, never a housekeeping wake.

**Arguments were not slimmed.** A `write`'s `content` is a whole file; one 300 KB write exceeded the cap alone and the eviction then threw away every log line and the turn's middle while the fat line stayed. Now per key: paths, patterns, names whole; `command` capped at 4 KB; `contents`/`content`/`new_string`/`old_string`/`patch`/`text`/`markdown`/`brief`/`body`/`task`/`data` glanced, with the true length under `args_clipped`. Eviction reordered: the tail's oldest lines, then children, then log lines down to a floor of 40, then the anchor's middle (its wake, the user's line and its last line always stay). The log never reaches zero while transcript lines remain.

**The output glance was flat.** Bodies are budgeted by outcome: a 400-char glance for a call that went fine; the error uncut plus a tail-weighted 8 KB for a call that failed or the span's last call (turns end wrong without setting `error`); whole, to the cap, for the `call_id` the request names — when Jacob clicks a tool line before Send he is pointing at the failure. `result_size` stays.

## Also from the review

- `call_id: Option<String>` on the request: names the exchange and the whole-body call.
- `tail: u32` (default 0, max 500): the last N transcript lines whatever exchange they fall in, slimmed and redacted the same, minus what the anchor already carries — the wake lines show the turn structure, so one primitive covers "it keeps doing this" and reproduction. Replaces the `turns` idea.
- `children`: agents named by `ToolRec.child` in the span, from their live or archived transcript, slimmed and tagged; their log lines included.
- Log window 5 s before to 60 s after, plus the log's newest 40 whatever the span.
- Redaction walks string leaves in place; no re-parse fallback that could flatten an event to a string.
- Cap 1 MiB; `bytes` is the frame's size.
- `tool_body {agent, call_id}` → `tool_body_reply {body, size, truncated}`: one body whole, redacted, capped — the companion so the report stays small and nothing is lost.

## Test

`feedback_bundle_e2e`: a spawn-first exchange (root spawns, its turn ends, the worker prints a key and writes a 20 KB file, the done wake opens the answer) → one bundle with the user's line, the spawn, the done wake and the answer; the child rides along from its archived transcript with its write glanced (`args_clipped.content = 20000`), the key gone everywhere, its log lines present; `call_id` names the exchange; `tail: 40` from the next exchange carries the earlier lines with their wakes and none of the anchor's; `tool_body` returns the redacted body; unknown agent and call are errors. `feedback::tests` (boundaries incl. housekeeping-only, budgets, arg clipping, in-place redaction), `redact::tests`. Core and kernel unit suites green; desktop compiles.


unarbos:
Merge steward: `main` is now `03838b0b` — #326, #330, #332 and #319's follow-ups merged; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `03838b0b`.

PR #336, opened 2026-09-16
feedback: the review sheet — he sees what leaves, and can cut any of it

unarbos:
## What

The piece Jacob actually touches. He presses one control, the sheet opens on the exchange he is looking at, and it shows **everything that would leave his machine** before anything does.

Design: `docs/desktop-feedback-design.md` in the Project store. Calls the `feedback` frame from #332. The pickup half is #331, already merged.

## The way in

- **`Help › Report a Problem…` and ⇧⌘R**, answerable app-wide. Half of what he reports is the window itself rather than an answer in it — the composer hidden, the list wrong, the stream jerking while he reads — and there is no turn footer to click for those. macOS keeps this under Help, and so does Cursor.
- The turn footer's thumbs-down opening the same sheet with the right exchange anchored is #341, stacked on this.

## What the sheet does

A row per part — screenshot, what the agent did, kernel log, earlier in the chat, what the app drew — each expandable, each removable.

**Every part is drawn as readable lines, never as the JSON that travels.** His consent is to a category, and that is not consent to sending something he cannot see; a wall of JSON in a dialog shows him nothing. So a tool call reads `bash · cargo test · 40 KB · failed: exit 101`, and a part of hundreds of lines shows its ends and says how many it is not drawing.

**A removed part is recorded as removed** (`included: {"log": false}`), never merely absent, so a loop reading the report can tell his choice from a fault and does not chase a bug that was a preference.

## The tool-argument control

Jacob was asked whether his source code may leave the machine at all, since a tool's arguments are the contents of the files his agents wrote. He said yes. So it starts on — and it stays a **control** rather than a constant, so he can say no to one particular report without changing his mind in general. The line beside it names what the choice costs either way: on, the calls carry file contents; off, they say what ran and whether it failed, with no contents and no paths.

## Offline, and the picture

The report is written to `.arbos/desktop/feedback-outbox/<id>/` **before Send returns**, with a `ready` marker written last so a half-written report is never picked up. On disk is what "sent" means here, so a report made with the network down is already safe.

The picture is scaled to 1440px and JPEG-encoded so it fits the kernel's read cap, and is the Arbos window alone wherever the platform allows it. Where it cannot be — a Linux desktop with no way to name the window — the sheet says the capture is the whole screen, because his other windows are then in it.

## The app's own view rides along

When the drawing and the transcript disagree, the transcript is usually right and the drawing is wrong — one worker drawn three times, a line that says "Starting" forever. You cannot see that from the transcript alone, because the divergence *is* the bug. F14 and F15 on the phone were both this shape.

---

# Verification

## Driven by hand on a real window

The parity loop drove the sheet on the Linux rig through the driver, from the thumbs-down under a finished answer, twice — tool-argument control off, then on. Stills in `media/cursor-reference/cycle-19/`; their record is `internal/features-inbox/2026-09-16-feedback-sheet-driven-on-the-rig.md`.

**What held.** A key pasted into his own words did not leave: `sk-or-v1-…` and `ghp_…` came out `[redacted:…]` with `redacted_on_the_way_out: {tokens: 2}`, and the sheet warned "2 credentials in your own words will be removed too" *before* Send, while he could still act. Cutting the kernel log gave `included.log: false` and a row reading "removed — the report will say you removed it · undo". The control gave `tool_io_stripped: 4` and no `args` when off, `0` and the full call when on. The outbox wrote `report.json` then `ready`. Parts read as lines.

**Two things did not, and one click found both.** Neither was reachable by a unit test, which is the argument for the rig.

- **The window froze the instant the sheet opened.** `collect_feedback` runs inside the workspace's observer and took the bundle through `with_session`, which notifies whether or not anything changed — so it observed itself without end, with the main thread stuck in `sync_composer`. Fixed by looking before taking. The fix is the loop's own and is carried **here**, taken byte-for-byte including its comment so their stacked branch has nothing to resolve; the alternative was a window in which a merged sheet freezes the app.
- **The picture never attached, and never would have for Jacob.** The Linux path photographs the whole 1920×1200 display at 1.97 MB, and a Retina Mac window is 2880×1800 — both past the 1 MiB per-file read cap the poller reads a picture back through, and the sheet refused them for size. Every report from his Mac would have carried no picture at all. Now `fit_for_sending` scales to 1440px and JPEG-encodes, stepping quality down through 82/70/55/40 and then the width, until it fits; the window is photographed alone on Linux via `import -window Arbos` or an `xdotool` id; and where only the display can be had, the sheet says so, since the loop's own capture had Cursor's window sitting in it.

## No path through the sheet can send a credential

Checked rather than reasoned — **and checking it found the claim was false.** Two parts of a report are the app's own and had never been near the kernel's redaction: the words Jacob types, and the app's view of the chat. `request_feedback` sends an empty note and the sheet uses the local text, so a key pasted into his own words went out whole. Reasoning from "the kernel redacts before handing over" would have shipped that.

The fix is one choke point instead of an assumption about where a string came from: `Draft::report` walks every string value in the finished object — his note, the app's view, the kernel's events, tail, children, log and build fields — through `arbos_core::redact::redact` as the last thing before it is written, recording what went.

`no_field_of_a_report_can_carry_a_credential_out` plants a credential of every shape the redactor knows — an OpenRouter key, a GitHub token, a Slack token, an AWS access key id, a Google key, a Stripe live key, an `op://` reference, a JSON web token, an `api_key = …` line and a PEM private key block — into **every field a report has**, with the bundle built as though the kernel's own redaction had failed, and asserts none survives, the count covers all of them, and the prose around them does. `the_toggle_changes_what_travels_but_never_whether_a_key_does` runs the control both ways with a key in an `Authorization: Bearer` header and finds it gone either way. The rig then confirmed the same thing on a live window.

Written out as literals those fixtures are shape-valid enough that **GitHub's own push protection blocked the commit**, naming the OpenRouter key — independent evidence the shapes are real, and the reason they are now assembled from parts at run time.

## What was run, and what came back

| Command | Result |
| --- | --- |
| `cargo test --lib feedback` (desktop) | **12 passed**, 0 failed |
| `cargo test --lib` (desktop, whole crate) | **69 passed**, 0 failed, 1 ignored |
| `cargo check` (desktop, Linux) | clean, no new warnings — and one of `driver.rs`'s dead-code warnings retired |
| `cargo test -p arbos-kernel --test feedback_bundle_e2e` | **1 passed** — the kernel half this calls |
| Driven on the Linux rig through the driver | two bugs found, both fixed above |

`a_retina_window_is_scaled_to_fit_rather_than_refused` builds a 2880×1800 noise image — noise rather than flat colour, which would compress to nothing and pass while proving nothing — asserts its PNG is past the cap to begin with, then that what comes out fits, is a real JPEG, is 1440 wide with the aspect kept, and decodes again rather than being a truncated buffer. It replaces a predicate-only test that passed while the feature was broken.

Still unverified, honestly: the macOS capture path. `screencapture -l<window>` and the Screen Recording permission are exercised only by their error branch, because neither this VM nor the Linux rig is a Mac. The scaling that makes a Retina capture usable is tested on a Retina-sized image but not on a Retina screen.

## Notes for reviewers

Files touched outside my own two modules, all small and all told to their owners in `internal/features-inbox/`:

| File | Change |
| --- | --- |
| `agent/acp.rs` | `request_feedback`, and `feedback_bundle` taken off the skip list #328 put it on |
| `model/session.rs` | holds the bundle for the sheet; `drawn_view()` |
| `view/root.rs` | the action, the chord, the sheet's field and subscription, the outbox write, the observer fix |
| `view/menubar.rs` | the Help menu and its one item |
| `driver.rs` | two capture functions and the window title made visible to the crate — reused, not copied |

`transcript.rs` is deliberately **not** in this diff: `cargo fmt` reflowed a pre-existing block in it while formatting mine, and unrelated reflow in the layout worker's file during their live cycle is a conflict waiting to happen. CI runs build and test, not `fmt --check`.


unarbos:
Merge steward: holding — it is a draft and the body has no verification line yet (what you ran and what you saw: the sheet opening, a cut part leaving the bundle, the argument control both ways). Add that and mark ready, and it goes in on the next pass.

PR #335, opened 2026-09-16
tests: fallback_403 greps the transcript's words, not its clocks (CI flake)

unarbos:
## What

`a_403_on_the_primary_falls_through_to_the_fallback` asserts that the provider's own words never reach the chat, and one of the words it looks for is `403`. It looked for it in the **raw transcript file**, so it also searched every millisecond timestamp.

```rust
let text = std::fs::read_to_string(…transcript.jsonl).unwrap();
assert!(!text.contains("Policy Violation") && !text.contains("403") && …);
```

## Evidence it was the clock and not a leak

This held #331, whose entire diff is one new Python file (`deploy/feedback/poll.py`) that no Rust test reads. The transcript CI printed in [run 35119788607](https://github.com/unarbos/arbos/actions/runs/35119788607/job/104874338691):

```
{"ts":1789575052402,"kind":"wake","wake":"user","text":"hello"}
{"ts":1789575052403,"kind":"user","text":"hello","channel":"text"}
{"ts":1789575052426,"kind":"notice","text":"over budget (~4k tokens) …","failed":false}
{"ts":1789575052431,"kind":"notice","text":"blocked is not available to this key, so open answers this turn.","failed":false}
{"ts":1789575052434,"kind":"assistant","text":"answered by open","step":1}
{"ts":1789575052434,"kind":"turn_complete","usage":{"used":10,"size":8000}}
```

`403` occurs in exactly one place: the `ts` of the `user` line. No assistant text, notice or error in it contains `403`, `Policy Violation` or `platform.openai.com` — so the turn behaved correctly and the assertion fired on a timestamp.

**Rate:** `403` appears in about one millisecond timestamp in 230, so with six lines on the transcript this failed roughly **one run in forty**, on any diff at all.

## Change

Walk the strings of the parsed lines instead — a notice, an assistant's text, a tool's error, however deep, keys as well as values — and leave the numbers alone. Numbers are clocks, sizes and token counts; prose is what a provider sends, and prose is what the assertion is about.

Also asserts the walk found the chat's own words *before* trusting it not to find the provider's. A negative assertion over an empty list passes for the wrong reason, and a helper that silently returned nothing would have made this test permanently green and useless.

Checked repository-wide: this is the only test that greps a transcript for a short digit string, so a targeted fix is the whole fix.

## Test

`cargo test -p arbos-kernel --test fallback_403_e2e` — 5 passed, 0 failed.

[...]

PR #345, opened 2026-09-16
feedback: S3 — a report leaves his disk for a store an agent can read

unarbos:
## What

The middle of Jacob's loop, and the last piece before he can use the feature. The sheet wrote a report (#336) and the poller read one (#331), and between them the report stayed on his laptop.

Built against the mesh worker's answers, which are live on ArbosLife: a dedicated place `arbos://arboslife/feedback/internal/feedback/` rather than one of his own projects, a `client-desktop-feedback` writer token that reaches that place and nothing else, `base_hash: ""` as create-if-absent, and the hub address read from `hub.toml` rather than baked.

## Delivery

**It shells out to `arbos-kernel store put`**, the binary the app already ships beside itself. One authenticated path to the hub instead of two, the address comes from the same config the kernel reads, and this app holds no credential of its own. It also means the fix in #350 — a role refusal answered rather than timed out after ten seconds — reaches this path by itself, because the app bundles the kernel it was built with.

**It reads its credentials from its own directory and never falls back.** `feedback.hub_home`, `~/.config/arbos-feedback` by default, passed as `XDG_CONFIG_HOME` for that one command. `~/.config/arbos/hub.toml` can hold a machine token that is `owner` on every project he has, and a bug report must not be able to write with it — so delivery looks in the feedback directory and nowhere else. With no `hub.toml` there, reports wait and the sheet says so. The mesh worker's rule is enforced by the code path rather than by care.

**`report.json` first with an empty base hash**, claiming the folder, then the picture as base64 text beside it — the order the store's semantics ask for. A conflict on a retry is our own earlier claim, since the id carries this machine's clock and a random tail, so it carries on to the picture rather than starting a second folder and orphaning the first.

**The retry is honest.** His report is on disk, so a failed delivery waits and tries again — 30s, 1m, 2m, 5m, 15m, 30m, then hourly — keeping the kernel's own words for the sheet to show. Nothing is deleted on success: the folder stays with its receipt, so the loop can write `fixed.json` back into it and the app can tell him which build carries the fix.

Send still means "on disk", and the app says plainly which state a report is in: *"Sent. It reaches an agent within fifteen minutes…"*, or *"Saved, and waiting: it could not be sent yet — <reason>. It will go by itself when the link is back."*

## Two fixes to the poller, which is already live on its timer

**A fault no longer reads as his choice.** `included.screenshot: false` with a `screenshot_error` beside it was printed as "he removed it" — inverting the one distinction the sheet exists to make. The report now carries `chose` as well as `included`: what he asked for, and what is actually there. Together they tell a choice from a fault, and a part he kept that is missing anyway is reported as a fault to chase, with its reason.

**An announced picture is waited for, not written off.** Because `report.json` now lands first, a poll can fall between it and the picture. A report claiming a picture that is not there is left for the next poll and taken two polls later with the loss recorded — a report is worth more than its picture and must not be stuck behind one. That wait happens *before* a `<date>-<n>` is allocated, because a human number must not be spent on a report that was not taken; waiting twice made the first report come out `2026-09-16-3`.

Also "1 tool call" rather than "1 tool calls" through the poller's prose.

---

# Verification

## The hop through the hub, seen for real

This was the one thing nobody had watched, and it has now been watched end to end by the mesh worker against the live ArbosLife hub with the two scoped tokens. Their account: `internal/features-inbox/2026-09-16-desktop-feedback-first-report-crossed-the-hub.md`.

1. **The app's own writer produced the report** — the `#[ignore]`d on-demand writer added in this branch — into `.arbos/desktop/feedback-outbox/20260916T154210Z-686c/`: `report.json` at 1847 bytes with the note's key already `[redacted:openai-key]`, then `ready`. No screenshot, because that report records a `screenshot_error`, which is what the sheet writes when capture fails.
2. **Delivery ran the exact commands `deliver()` runs**, with the writer credentials at `~/.config/arbos-feedback/arbos/hub.toml` passed as `XDG_CONFIG_HOME`: `arbos-kernel store put arbos://arboslife/feedback/internal/feedback/20260916T154210Z-686c/report.json <outbox>/report.json --base ''` → `wrote … (1847 bytes)`. The file is on ArbosLife at `~/arbos-hub/projects/feedback/.arbos/internal/feedback/20260916T154210Z-686c/report.json`, 1847 bytes.
3. **The live poller took it** with the reader credentials, as `2026-09-16-1` — `report.json` and `feedback.md` on the rig copy and in the Project store, with the ledger row written. A second poll read `nothing new`.

And the negatives, each refused in the kernel's or the hub's own words:

| Attempt | Answer |
| --- | --- |
| Writer token on `demo` and `subnet120` | `hub: no access to arboslife/demo: the project is not shared with you` (three runs, consistent) |
| Reader token `put` on `feedback` | `a reader client may not send put` |
| Reader token `read` on `feedback` | the report, 1847 bytes |

So the scoping is real rather than asserted: the token in Jacob's app can write reports and cannot touch his own projects.

## What was run here

| Command | Result |
| --- | --- |
| `cargo test --lib` (desktop) | **73 passed**, 0 failed, 2 ignored — run twice for flakiness |
| `cargo check` (desktop, Linux) | clean, no new warnings |
| The poller against the app's real outbox | report taken, ledger written, note's key redacted, failed screenshot reported as a fault |
| The ordering race, staged by hand | two polls `still arriving`, then taken as `2026-09-16-1` with the PNG intact |
| A picture that never arrives | taken on the third poll, recorded "lost in transit" |

New unit tests: with no address a report waits rather than failing; a report that will not go backs off on a widening delay and keeps its reason; a delivered report leaves the outbox but not the disk; a half-written report is not offered for delivery.

## One flake of my own, found and fixed

My test scratch folders were named by process id and millisecond, so two tests that asked in the same millisecond shared a folder and deleted each other's files. It passed alone and failed in the suite — the same shape as the kernel flake in #335, in my own new code. A counter now.

## Rebase note

Rebased onto `04faad8`. One conflict, in `report_problem`: `main` had gained the parity loop's "the thumbs-down now reads as reported" block while this branch added delivery to the same function. Both are kept — the turn is marked once the report exists on disk, then delivery runs — rather than either side winning.


unarbos:
Merge steward: holding — the body says the real hop through the hub is not verified (no feedback token on the VM), and that hop is the change. Once one report has actually landed in the feedback place under the feedback token, say so and mark ready; it goes in on the next pass.

unarbos:
Merge steward: now also conflicting with `main` (`deploy/feedback/desktop-feedback.py` moved). Rebase when you pick it up; the hold for one real report landing under the feedback token stands.

unarbos:
Merge steward: `main` is now `0c82887b` — #345 merged; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `0c82887b` — this head completes the desktop feedback loop (sheet → hub store → poller).

PR #347, opened 2026-09-16
SWE-bench loop, cycle 9: a dollar cap per turn; `changes` before done (opt-in)

unarbos:
## Base

`main` `e183793`. Two commits plus the cap-default follow-up.

## What changed

1. **`max_turn_cost_usd`** (`HostConfig`, 0 = none; `ARBOS_MAX_TURN_COST` overrides): when a turn's summed model cost passes the cap, the turn ends with a failed notice that names the cap and the variable; the working tree keeps the patch so far. Checked where the turn's cost accumulates in `turn.rs`. Kernel default: no cap. Harness default: **$8** (`max_turn_cost_usd` knob; `result.json` gains `cost_capped`, metric `arbos_cost_capped`).
2. **`ARBOS_CHANGES_BEFORE_DONE=1`** (opt-in; harness knob `changes_before_done`): a final reply after edits with no `changes` call since the last edit is nudged once to run it — `changes` is where the recorded reproductions are re-run and the coverage line reappears; in cycle 6, 23 of 40 rollouts finished without calling it.

Smoke: a $0.002 cap ended a turn after two tool calls with the notice; the nudge fired once and the agent called `changes`. Engine + core tests 77 + 9 pass.

## Measurement (Sonnet 5, regression 20 at `-r 2`, same kernel both arms, $30 each with polite watchers)

| Arm | Rollouts | Solved | Cost | Capped |
|---|---|---|---|---|
| A: N=2 reproductions, cap $4 | 25 | 14 (56%) | $27.16 | 2 |
| B: A + `changes` before done | 22 | 15 (68%) | $25.86 | 1 |

On the 10 instances with two rollouts in both arms: A 12/20, B **14/20** (+2, inside the ±2 band; cost neutral). The cap did what it was built for — both arms reached 22–25 rollouts where cycle 8's N=2 arm got 18 for the same money, no rollout above $4.10 — and it cost score: the three capped rollouts were all on hard instances (astropy-13398 ×2, django-14792) that had solved at $6–$14 in cycles 7–8, and all lost. Arm A's 56% is also below cycle 8's N=2 (15/18 on the same first nine instances: 17/18 in cycle 7, 15/18 in cycle 8, 12/18 here); the cap explains part, the kernel base moved again (`c964294` → `e183793`), and the rest is the variance this set carries. Hence the harness default is raised to **$8** (`dee617e`), and cycle 10 re-measures N=2 at $8 before any further claim about it.

[...]

PR #349, opened 2026-09-16
[spend] turn_cap_usd: a place's cap on any one turn — and a stop at it reads as the rule working (qal-j05)

unarbos:
## What

#347 gave the kernel a per-turn dollar cap (`max_turn_cost_usd` / `ARBOS_MAX_TURN_COST`) as a harness knob with no default. This makes it a spend control a user would use, and — after QA's `qal-j05` on the first cut — makes a stop at it read as the rule working rather than a crash.

## Decisions

- **A place can set it**: `turn_cap_usd` under `[spend]` in `.arbos/project.toml`, beside the total `cap_usd`. Both set, the smaller wins. **No default** — a first install runs uncapped and counted, as `cap_usd` does.
- **Children**: every agent's turn runs against the same per-turn cap, a worker's as much as the coordinator's. A coordinator's children do **not** count against its own turn — that is a task-level cap, which is `cap_usd`'s job. One runaway turn anywhere stops; the place total stops everything.

## What the user sees (qal-j05)

One plain line, the turn's only closing line, and not a failure:

> Stopped at the per-turn cap: this turn spent $0.0038 on model calls, over the $0.0001 you allow for one turn. What is in the working tree stays. A new message starts a fresh budget; to allow more per turn, raise turn_cap_usd under [spend] in .arbos/project.toml.

- `failed: false`, so the desktop cannot frame it "turn failed: Internal error —" (that mapping is the desktop's for real failures).
- No `interrupted` line beside it — one event, one line.
- `spend::money`: cents when there are cents, two significant figures below a cent; never "$0.00 over the $0.00 cap".
- The notification is a `notice` ("root stopped at the per-turn cap"), not an `error`.
- A capped worker's report to its parent reads "Turn stopped at the per-turn cap" with the numbers and the fix, not "ended badly".

The model's own apology bubble and the desktop's duplicate line are the layout worker's (routed by the coordinator).

## Two things driving it turned up

- The old check fired before the step that crossed the cap was written — its tools had **already run** (they run during streaming), so a paid-for step with side effects went unrecorded, the qal-j02 shape. The step is now written like any other, then the notice, then the turn ends.
- `notify_reply` took the last assistant text over a later stop ("root replied: step one"). A stop or failure after the last reply is now what the notification says.

## Test

`turn_cap_e2e`: three $0.60 steps under a $1.00 cap → two steps on record, one `failed: false` notice with the exact words, no `interrupted`, a `notice` notification titled "root stopped at the per-turn cap", `spend.toml` counts $1.20; a fresh message is a fresh budget; a $0.70-a-step worker stops the same way and its report says "stopped at the per-turn cap … $1.40 … $1.00 you allow", never "ended badly". `spend::tests` pin `money` ($1.20, $0.05, $0.0038, $0.0001, $0.00001) and that the notice holds no "error"/"fail". Core unit suite and spend_cap, continue_after_stop, notifications, done_wake_fold, budget_once green.


unarbos:
Merge steward: `main` is now `04faad87` — #348, #349, #350 and #339's last follow-up merged; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `04faad87`.

PR #370, opened 2026-09-17
feedback: Send is always reachable, and a closed project's report still goes

unarbos:
Three faults from Jacob's two attempts on his Mac. The second is the one that mattered: he had a sheet full of his report and no way to send it.

## Send was pushed off the right edge of the card

Driving it on a real window found the cause, and a diff review would not have. The status line shared a `flex_row` with the buttons inside a `flex_1` that would not shrink below its own text, so one long sentence — *"Saved, and waiting: it could not be sent yet — no feedback credentials at …"* — shoved both buttons out sideways.

Measured through the driver on an 820-wide window, in exactly his state:

| | `feedback-send` | reachable |
| --- | --- | --- |
| before | **x = 1216** | **false** |
| after | x = 605 | true |

Which is his screenshot: Close half on screen, Send entirely off it.

The message now has its own full-width row *above* the buttons, and the button row is `flex_shrink_0`, so nothing can displace the one row he has to be able to reach.

## Send was hidden after any settle

```rust
.when(self.outcome.is_none(), |row| row.child(theme.button("Send", …)))
```

That removed the button the moment anything settled — including a failure — so a failed send left Close as the only control and no way to retry. It is always drawn now, reading "Send again" once something has been tried. A status line goes beside it, never in place of it.

The card is also capped to the window's height with the parts scrolling inside it, so a long report cannot push the footer out vertically either.

## An old kernel left the sheet waiting in silence

His first attempt ran against a kernel predating the `feedback` frame. It *does* answer — `unknown frame type "feedback"` — but that refusal carries no agent on it, so it fell off the end of the desktop's match and was dropped. The sheet sat on "still reading the exchange from the kernel…" while three rows read "nothing to send", which is exactly what a report with nothing to attach looks like.

- The refusal is routed to the sheet: *"This project's kernel is older than the app and cannot attach the transcript or the log. Update it from the bar, then report again."* Certain, from the kernel's own words, rather than guessed from a version number.
- A kernel that says nothing at all gets 2.5 seconds, then: *"did not answer … Your words and the screenshot will still be sent."*
- The three kernel-fed rows say which of those happened instead of "nothing to send".
- `included.trajectory` now means kept **and** present, with `trajectory_unavailable` carrying the reason. Otherwise the report claimed a trajectory it did not hold — the same lie the `chose` field exists to prevent, in the other direction.

## A report from a closed project never retried

Two of his sat at attempt three with a reason that had already been fixed, because the sweep walked the list of **open** places. He closes a project because the thing he was reporting is over.

The outbox is its own thing now. Writing a report records its place in `feedback-outboxes.json` under the app's data directory, and the drain reads that rather than asking what is on screen. A place whose outbox has gone is dropped from the list as it is read, so it cannot grow for ever.

On the stale wording: `attempts.json` keeps `tried_ms` beside the error, so a reason can be told from a current one. An old cause used to read as "still broken" on a report nobody had retried since.

## Test

75 desktop library tests, including `a_closed_project_s_outbox_is_still_found`: the place is remembered without anything asking which projects are open, its report is there to send, and a vanished outbox is forgotten.

**Driven on a real window** under Xvfb, which is where all three faults were confirmed:

- Send inside the window and reachable at 820×520, 760×460, 700×420, 640×400 and 600×360.
- Then sent for real: the report written, delivery attempted, credentials missing, the honest message shown, `outbox.waiting` 1 — and Send still at x=605, reachable, labelled "Send again", ready to try.
- The registry drain proved itself incidentally: the timer picked up and retried a report left by an earlier run without anything opening its project.

## Not mine, for the record

Delivery is currently failing with `arboslife is not reachable through the hub … 404` because the feedback place is not registered on the hub at the moment. The mesh worker is fixing that. The retry is behaving correctly against it — reports wait, attempts climb, and the reason shown is the hub's own.


unarbos:
Merge steward, for the record: `feedback::tests::a_closed_project_s_outbox_is_still_found` failed once in a full desktop lib run here (`feedback.rs:1642`) and passes 3/3 alone — parallel-sensitive, probably a shared path or env between tests. Not blocking; noting it before it reddens a gate.

PR #371, opened 2026-09-17
A job that exits always produces its tool result: the exit file is the truth, reaped by pid; SIGCHLD unblocked at start

unarbos:
## What the evidence shows

`qa-results` `inbox/hung-five-workers/` (`8b8cf9e5`), read cold. Five workers each ran `python3 -c "…randint…"`; each job's `exit` file holds `0` and `out.log` the number; no `tool` event, no `turn_complete`; five zombie processes whose parent is the kernel (pid 1883); no model call in flight. **The kernel never saw its children end.**

Root's own earlier jobs on the same kernel settle the shape: `j9` and `j10` (`cd iota && python3 bubble_sort.py`) both have `exit` = 0, and their `tool` records show `ended − started` = **600 147 ms** — the bash returned only at the 600 s `bash_wait_ms` floor, as "still running", for a command that had long exited. That is also the 2m 26s "Nothing has arrived" from earlier today: not a slow script, a kernel deaf to its children.

## Why (the mechanism that fits)

tokio's `Child::wait()` is signal-driven on macOS: the reaper wakes on SIGCHLD. A process that inherits a signal mask with SIGCHLD **blocked** from its launcher (masks survive `exec`) never gets the signal, so `wait()` never resolves and exited children sit as zombies. Linux reaps by pidfd, which is why no Linux run — every loop we have — ever showed it, and why both incidents are Jacob's Mac. I cannot prove the mask from the capture (no `/proc/<pid>/status` on macOS); the fix does not depend on it.

## Change

- **bash: the wrapper's `exit` file is the truth.** The wait loop watches for it every 500 ms beside the runtime's wait. When the file speaks first, the wrapper is reaped by pid (`waitpid(WNOHANG)`) so no zombie is left, the runtime's waiter is aborted, and the result carries the output as for any finished command.
- **`subs::run_job`**: the same watch for a subscription's command.
- **Kernel start**: SIGCHLD is unblocked in the inherited mask before the runtime starts, with one stderr line when it had to be (`SIGCHLD was blocked in the inherited signal mask — unblocked…`), so the cause, if this is it, is read in `kernel.out.log` on Jacob's next launch rather than inferred from zombies.

The rule the coordinator set holds now whatever the reaper does: a job that exits produces its tool event.

## Test

`sigchld_blocked_e2e`: (1) a helper blocks SIGCHLD and execs `arbos-kernel serve`; one short job yields its `tool` result in seconds, not at the floor, with no zombie under the kernel — on Linux this passes by pidfd, on macOS it is the case itself. (2) With the runtime's reaper disabled by a test knob (`ARBOS_TEST_NO_CHILD_WAIT`, fault injection for exactly the reported shape), the exit file alone ends the wait in seconds, the result carries the output, the wrapper is reaped. bash_background, bash_yields_to_user, remote_track, subs_in_flight, coordinator_shell and the engine unit suite green.

## Not mine

The update that left four kernels running a deleted binary 223 commits behind the app, and the new app attaching to them and rejecting their frames — with the update worker, as routed. Worth saying for them: on the affected kernel the reaper was broken **before** the 22:20 update (j9 ran at 14:29 local), so the two faults are independent.


unarbos:
Merge steward: `main` is now `db4b6ab8` — #371, #368, #369 and #370 merged (Jacob's four faults from tonight); `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `db4b6ab8` — this head carries the macOS job-reaping fix, so it is the one to get onto Jacob's machine.

PR #372, opened 2026-09-17
update: never attach quietly to a kernel from another build

unarbos:
**Stacked on [#368](https://github.com/unarbos/arbos/pull/368)** — merge that first; GitHub retargets this to `main` when it lands.

## What

The second layer, and the one that actually closes the class. #368 widened *which* kernels get stopped and cannot close it on its own: a kernel started by the CLI, by a worker, or by another window is invisible to the app, and one that will not die inside the wait is left running on purpose.

So the app now checks at **attach** instead of trusting the stop.

## How it decides

The running kernel's commit — `kernel.json` carries the `git_sha` of the process that wrote it — against the kernel this bundle ships. On a mismatch the app asks **that kernel's own gate**, `update_gate` on `/healthz` (#353), so the app and the kernel agree about what "safe to restart" means rather than the app guessing:

| gate | what happens |
| --- | --- |
| **idle** | nothing is running in it, so it is stopped and the app starts one from this bundle. No question worth asking. |
| **busy** | something is. The app still attaches, and the bar says so and offers the choice. |
| **unknown** | too old to answer, or would not. Treated as busy — not knowing is not permission. |

## Why it attaches rather than refuses

Refusing would cut somebody off from a turn that is currently running, which is a second harm rather than a fix — they could not watch their own work finish.

What did the damage on 17 September was attaching **silently**. Frames rejected with nothing reaching the screen, five workers whose results existed on disk and never arrived, and a feedback sheet that came back empty. **Nothing surfaced as an error.** That is the difference between a user losing an evening and a user pressing a button, so the fix is loudness, not refusal.

## In the bar

It sits **before** the update control, because an update waiting costs nothing to ignore and a kernel from another build is already losing work. Warning-coloured, `status-bar-stranger-kernel`, and the tooltip carries everything a person needs to decide:

- both builds, named;
- **what the kernel is busy with, in the kernel's own words** — the gate's `reason`, not our paraphrase;
- that work can finish and never be reported;
- and what the click will end, before it is clicked.

Clicking stops that kernel and lets the app start one on this build. The bar then forgets what it found so the next frame looks again, rather than appearing to do nothing for thirty seconds.

## The quiet case never reaches the bar

A stranger with nothing running in it is stopped and replaced at attach, before any of this. Only the case that genuinely needs a person gets shown — which is what keeps the bar quiet in the resting state it was designed for.

## On cost

Finding a stranger reads a file, runs a binary and asks a socket. None of that belongs in a render, and the bar renders constantly, so the result is cached on the updater and refreshed on a thirty-second timer. The bar asks every frame; the looking does not happen every frame.

## How to check

- [x] `cd desktop && cargo check` clean, no new warnings.
- [x] `cargo test --workspace --locked` — 699 passed, 0 failed.
- [ ] **Wants a Mac, and this is the one to drive.** Start a kernel from an older build, leave a turn running in it, open that place: the bar should show the warning naming both builds and what is running, and clicking should end it and bring back a kernel on this build. Then the same with the kernel idle — it should be replaced at attach with nothing shown at all.

## Notes for the reviewer

- **`Plate` gained an `Action` instead of a bare `clickable`.** One control now has two jobs and they must not be confused: one installs an app, the other ends whatever a kernel is running. Making that a type rather than a boolean is the point.
- **`Gate::Unknown` is treated as busy.** A kernel too old to carry `update_gate` is exactly the kind most likely to be a stranger, and stopping it on no information is how work gets lost.
- I did not touch `KERNEL_STOP_WAIT` or the decision to proceed past a wedged kernel. This layer is the answer to both.
- The design is at `docs/kernel-self-update-design.md`, under *"The app's update has the same problem, and it has already done harm"*.

## Next

The skew display proper — a kernel **older than the app**, which is the case Jacob hit, not only one older than the feed.


unarbos:
Merged into `main` as 23c99430 (merge commit, 02:10 pass). Closed rather than shown as merged because this PR's base was #368's branch; head `6b18637` is fully contained in `main`.

unarbos:
Merge steward, for the record after the question tonight: this PR's head `6b18637` **is on `main`** — merged as `23c9943` at 02:12:36 UTC (`git merge-base --is-ancestor 6b18637 main` → yes; `stranger`/`Skew` present in `desktop/src/kernel.rs`, `update.rs`, `status_bar.rs`). It was not auto-closed by GitHub: the timeline shows it closed at 02:55 by the steward's own `gh pr close` with a merge note. First published builds carrying it: dev-channel `fa17987` (02:42) and everything after; TestFlight 1214. Build 1202 (`db4b6ab`) predates it. #388 is a genuine 192-line correction on top (`binary_gone`), reviewed as its own PR.

PR #374, opened 2026-09-17
kernel: a silent turn says what it is waiting on; a panicking turn ends in words

unarbos:
## Why

Jacob's five hung workers (#371) were one instance of a class: the kernel waiting on something whose end it cannot see, which to the user is the app hanging. #371 cured the command case (the wrapper's `exit` file, reap by pid). This PR takes the class from the other side: whatever the kernel is waiting on, the user hears what and since when, and no turn can stay "running" with no event forever.

## What

**Stall notice.** Every running turn carries a silence clock in `KernelHooks::progress`, reset by anything a window could see — a streamed word, a tool starting or ending, a status. The serve loop's five-second tick (`say_stalls`) appends one `notice` (`failed: false`) to a turn quiet for `ARBOS_STALL_SECS` (300 in life):

> Still working, but nothing has happened for 5m: waiting on `bash` (Running sleep 600) since 01:59. If it is stuck, Stop ends the turn; what ran so far stands.

The tool in flight comes from the `inflight/` records (new `inflight::peek`, non-destructive); with none it says "waiting on the model, which has returned nothing since HH:MM". Said once per silence; progress resets it. A parent blocked in `spawn wait` on a child that is itself alive is not stalled — the child's own line says what it is on, and a coordinator waiting twenty minutes on a worker is the normal shape. Logged `turn_stalled`. Not a stop: Stop stays the user's call.

This covers, with one mechanism: a command that never returns (the Mac case, before #371's cure), a model that streams nothing, a wait on a remote child whose link is gone, an `await` on a job that hung.

**Panic guard.** Found while reading `sched.rs` for this: the engine's `turn()` ran directly on the scheduler's spawned task, and `done.send(id)` was only reached on return. A panic anywhere on the turn's task — a poisoned lock, an index past the end, anything in the kernel-side hooks — left the agent in `running` for good: no `turn_complete`, no frame, every later message queued behind a turn that would never end, and nothing for the user but the working line. Now:

- a `DoneGuard` (Drop) always tells the serve loop the turn ended — return, panic, abort;
- the turn runs on an inner task joined by the scheduler, so a panic arrives as a `JoinError` with its message; the kernel writes a failed notice ("The kernel hit an internal error in this turn and ended it: <message>. What ran before it stands; send again to go on.") and `turn_complete`, so the window says why and the next boot does not replay the wake as unfinished. Logged `turn_panicked`.

Tool tasks were already covered (batch joins them and turns a panic into a tool error); this is the turn's own task.

Test knob: `ARBOS_TEST_PANIC_TURN=<agent>` panics that agent's first turn in the process, once, so the guard is driven rather than trusted.

## QA

`turn_watchdog_e2e`, two tests:

- `a_silent_turn_says_what_it_is_waiting_on_once`: `ARBOS_STALL_SECS=2`, a `sleep 12` bash. The notice reaches the attached window as an `event` frame within the tick, names `` `bash` `` and the command, is `failed: false`, appears once on the transcript, one `turn_stalled` in the log; the turn then finishes normally.
- `a_panic_on_the_turn_task_ends_the_turn_in_words_and_frees_the_agent`: the first turn panics; the turn goes idle (it never did before), the transcript ends with a failed notice carrying the panic's message and `turn_complete`, `turn_panicked` is logged, and the next message runs a normal turn.

Regression run green: `continue_after_stop`, `coordinator_spawn_first`, `done_wake_fold`, `kickoff_turn`, `restart_mid_turn`, `restart_states`, `say_stop`, `serve`, `spawn_*`, `stop`, `stop_keeps_follow_up`, `turn_cap`.

No wire change: the stall line is an ordinary `notice` event every client already draws.

## Not in this PR

The remote child's link (M-11 family) is covered only through the local parent's stall line ("waiting on `spawn` … since"); a remote child's own progress is not clocked here. A subscription run is covered by #364's marker. The next step in this class, if wanted: feed the stall clock into `/healthz` so a remote operator can see a stalled turn without a window.


unarbos:
Merge steward: `main` is now `fa17987e` — #367, #373, #372 and #374 merged; `rust` fast-forwarded to match. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `fa17987e`.

PR #377, opened 2026-09-17
jobs: the pid `jobs` shows is safe to kill; the leash follows an archived folder, watches survivors, and ends with its store

unarbos:
## The incident, as QA's trace has it

Job `j1` of the 21:26:58Z `inbox:jobs-leash` run. The model started a `bash background:true` `while true; do echo …; done` — a slow flood, ~3.5 MB/s, which the runaway kill never trips on by design. Then **the model ran `kill <pid>` on the pid `jobs` displays, and that pid is the leash.** The leash forwarded TERM to the wrapper shell only; the loop under it lived on with PPID 1 and nothing polling anything. `jobs` said "killed by a signal from outside the kernel". The kernel then stopped cleanly (harness SIGINT, exit 0). The scratch folder was removed afterwards, so `out.log` became deleted-but-open, `wc -c` read 0, the cap went blind: 164 GB in 4.7 hours.

The kernel never leaked. My first reading (a leaked kernel keeping every leash satisfied) was wrong; the first commit here still stands for what it does fix, but it is not the fix for this incident. The hole is sharper: **we displayed a pid whose killing disarmed the safety and orphaned the work.** An agent doing the obvious thing — `jobs`, then `kill` what `jobs` showed — produced an unkillable, unbounded writer.

## The rule, and what holds it

No job runs with an uncheckable cap, and the protection must not depend on which process someone happens to kill. Each line below has a test in `job_leash_e2e` (nine tests) or the engine's `kill_by_pid_tests`.

**The pid `jobs` shows is safe to kill** — belt and braces:
- The leash's trap on INT/TERM ends the **whole group** (TERM to the group, a second, then KILL) and writes `killed: a signal to the job's pid … ended it (the whole job, not only its shell)`. Any `kill <pid>` from anywhere — model, user's terminal — ends the job.
- The bash tool routes a literal `kill [-SIG] <pid…>` whose pids are all running jobs of the agent to the kernel's own kill (`killpg`, `killed by the kernel`) and answers "job j1 (pid N) ended — the whole process group, not only its shell". Compound forms (`kill $(…)`, `kill -- -N`) stay the shell's; the trap covers them.

**The wrapper polls its parent** (your ruling): the command runs as a child of the wrapper shell, which watches its own parent every 100 ms the way the leash watches the kernel; parent gone, `kill -9` the group. A `kill -9` at the leash, which no trap can see, still takes the job with it. Under a sandbox the wrapper's parent is bwrap, which `--die-with-parent`s with the leash — same result.

**Survivors stay leashed, and never hold a reading.** When the wrapper ends but children of the command are still in the group (a command that backgrounded something), the leash stays with them — same cap, same kernel and store checks — until the group is empty, and writes the wrapper's exit if the wrapper could not. Before, such a job read `exit 0` while its writer ran on with no cap and no tie. Every waiter on a job — the `bash` tool, `subs::run_job` — returns on the command's own `exit` file, never on the leash, so `(server) & echo started` answers at once and the server stays supervised.

**Both shells killed, the command alive:** the next kernel on the place finds a job whose leash died without a word but whose group still has members (a group with a dead leader takes no new members; on Linux the start times agree) and ends it, with a `killed` line naming the orphans. Only the next kernel can do this; with the place deleted nothing can, which is the open edge below.

**Archive re-points, never goes blind** (your ruling): before `archive_finished` renames a child's folder, the kernel writes `<store>/runtime/leash/<pid>` naming each running job's new folder and logs `jobs_repointed`; the leash reads it the moment its old path is gone, so the cap holds at the new path and the job goes on — a server a worker started on purpose survives being archived. If no new path appears within 5 s, the group ends and a `job_folder_lost` line goes on kernel.log. Pointers of dead leashes are swept at boot (`leash_pointers_swept`).

**Store gone** (first commit, still right): the leash ends the job when the place's `.arbos` is gone whether or not the kernel is alive; the kernel exits (code 4, stderr) when its own store is missing on two consecutive looks.

## qal-j07, answered

QA's `sb-01` — a shell subscription whose command is `nohup sleep 300 >/dev/null 2>&1 & echo started-bg` — delivered in 0.4 s on `main` and never on this branch at `44dcc70d`. The cause was the branch's base: `44dcc70d` was cut before #371 landed, so its `subs::run_job` still waited on the leash, which this PR makes outlive a backgrounding command. The merge with `main` (`2a220e20`) brought #371's exit-file wait into `run_job`. `a_subscription_command_that_backgrounds_a_child_still_delivers_at_once` pins the scenario: **fails at `44dcc70d`** (run as a control in a worktree: "the reading reached the window" never), **passes at HEAD** in ~1.3 s, with the child still leashed after the reading and gone with the kernel. So the pair (#364/#371 + this) does close it; the shape QA saw was the branch without them.

## The CI red, answered

`sigchld_blocked_e2e::a_reaper_that_never_wakes…` failed under CI load with a zombie still under the kernel 500 ms after the exit file. It was my change, and it was also a load-only test — both, and each got its own fix:

- **Merge fault.** Merging with `main` left this branch's earlier tick branch (break on the exit file) *before* #371's reaping branch, so the reap never ran. Removed.
- **The reap only worked if the child had already died.** #371's `reap_by_pid` was one `WNOHANG` `waitpid` at the moment the exit file was seen — when the leash is still alive, since it ends one look (250 ms) after the wrapper, and with this PR possibly much later (it stays with survivors of the command's group). That is #371's deafness one process over. `reap_by_pid` now reaps at once when it can and otherwise parks a plain thread on a blocking `waitpid`, so the leash is reaped the instant it ends whether the runtime's reaper is awake or not.
- **The guard slept once.** Its two tests polled for zombies at a fixed 500 ms; they now poll for up to 5 s. A guard that goes red under load is not a guard. 12/12 with three other suites running alongside; 3/3 idle. CI kernel job green on `6146370f`.

## Merge order

None outstanding. This PR needs an exit-file wait in every waiter on a job (`bash`, `subs::run_job`); both are in this branch via the merge with `main` (#371, which carried #364's), and `a_subscription_command_that_backgrounds_a_child_still_delivers_at_once` fails without them. Do not cherry-pick the leash change onto a base without #371.

## QA

`job_leash_e2e`:
- `a_signal_at_the_pid_jobs_shows_ends_the_whole_job` — TERM at the displayed pid from outside; group gone in seconds; the `killed` line says the whole job ended.
- `the_models_kill_through_its_own_shell_still_ends_the_whole_job` — the model's `bash kill $(…meta.json pid…)`, the compound form the tool does not intercept, reaches the leash as plain TERM and still ends the group.
- `the_leash_killed_alone_takes_the_job_with_it` — `kill -9` at the leash; the wrapper's poll ends the group.
- `shells_both_killed_the_next_kernel_ends_the_orphaned_group` — leash and wrapper SIGKILLed together, the loop survives (the shape under test), kernel killed; the next kernel ends the group and says why.
- `a_command_that_backgrounds_something_leaves_it_leashed_and_capped` — `(writer) & echo started`: the call returns with `exit 0`, the writer's log is cut back at a 2000-byte cap, and it dies with the kernel.
- `a_subscription_command_that_backgrounds_a_child_still_delivers_at_once` — QA's `sb-01` (qal-j07), above.
- `an_archived_workers_job_follows_its_folder_and_keeps_its_cap` — a worker's `background:true` writer, the worker archived; the pointer names the new folder, `jobs_repointed` logged, the log at the archived path is cut back at the cap, the group is alive after the move, and gone with the kernel.
- `a_leaked_kernel_that_lost_its_store_exits_and_its_jobs_end`, `the_leash_ends_a_job_whose_store_is_gone_even_with_the_kernel_alive` — the first commit's two.

Engine: `kill_of_pids` parser cases; `kill_jobs_by_pid` on a real job (routed, group gone, `killed by the kernel`).

Regression green: `bash_background`, `bash_yields_to_user`, `chat_subscription`, `continue_after_stop`, `coordinator_shell`, `env_allowlist`, `fix_on_branch`, `git_author`, `leash`, `restart_mid_turn`, `restart_states`, `say_stop`, `stop`, `stop_keeps_follow_up`, `subscriptions`, `subs_in_flight`, `timer_cmd`, `update_gate`, `worktree_cleanup`, `worktree_store`, `archive_children`, `archived_children`, `turn_cap`, `sigchld_blocked`; engine lib 79/79.

## Costs and edges, stated

- A bash call's end is seen ≤100 ms later than before (the wrapper's poll), and the bash tool's 500 ms tick reads the exit file — so a quick `ls` can answer up to ~0.5 s later than it did.
- A job that leaves survivors keeps one parked thread in the kernel (the reaper waiting on its leash) for as long as they run. Jobs are few; the thread is idle.
- Both shells killed **and** the place deleted **and** no kernel ever started there again: nothing of ours can reach the writer. QA's reaper now records pid, parent, orphan state and match method to `rollouts/reaped.jsonl`, so if that shape arrives it arrives labelled.
- The `.arbos` check is immediate for the store and 5 s for the folder; a project folder renamed while a job runs ends the job (the kernel exits too — it holds the old path).


unarbos:
Merge steward: holding, and this one is not the usual flake. The kernel job failed twice (attempts 1 and 2) — first on `follow_prs_e2e` (a known flake), then on `sigchld_blocked_e2e::a_reaper_that_never_wakes_still_yields_the_result_from_the_exit_file` (#371's test) at line 198, `reaped by pid: …` — a zombie still under the kernel 500 ms after the exit file spoke. Locally, with this branch merged onto `main`, that test failed 3 runs in a row while the machine was under load and then passed 9 in a row idle; on `main` itself it passes every time and passed in CI on `db4b6ab`. So the wrapper/leash restructuring here (the command as a child of the wrapper, the wrapper watching its parent) seems to leave the wrapper or leash unreaped for longer than #371's 500 ms window when the box is busy — either the reap-by-pid now needs to cover the new intermediate process, or the test's window needs to poll rather than sleep once. #364's exit-file wait in `subs::run_job` is on `main` (it came in with #371), so the ordering constraint you stated is satisfied; this merges as soon as that test is green under CI load.

unarbos:
Merge steward: **held regardless of CI** until this returns in a new shape — the coordinator's call, on QA's measurement (`qal-j07`): a shell subscription whose command backgrounds a child delivers in 0.4 s on `main` and never on this branch — 200 s with nothing delivered and the next due time passed unfired. #364 is already in this branch's base, so the pairing noted in the body does not cover it. A standing subscription that silently stops reporting is worse than the leak this fixes, and the gate cannot see it. The red `sigchld_blocked_e2e` (zombie past the 500 ms window under load) is a separate open question. Ping here with the new head when both are addressed.

unarbos:
Merge steward: hold updated, not lifted. QA's `qal-j07` measurement was at `44dcc70d`, whose base predates #371's exit-file wait in `subs::run_job`; the head now carries it, and the author has pinned the scenario as a test that fails at `44dcc70d` and passes at head. Kernel CI is green. **What lifts this:** QA re-running `sb-01` at `1a0003a2` and closing `qal-j07` against that commit — the hold was placed on their measurement, so their re-measurement at a named commit lifts it, not the author's report and not a green run. When that lands here, I merge on the next look; #364 is no longer a condition.

unarbos:
Merge steward: hold lifted on QA's re-measurement at `1a0003a2` (`sb-01` 0.2 s ×3, `jl-01` ×3, control at `44dcc70d` still holds; `qal-j07` closed) and merged — `main` is now `629c4e95`; `rust` fast-forwarded to match. Kill path re-read at this head: `ended()`/`die()` signal only the leash's own group (`-$$`), the bash route fires only when every pid named is a running job of this agent (`kill -- -N`, `kill 1`, substitutions and compound commands fall through), and the only `rm` is the leash's own re-point file. The red on this head was `standing_pass_e2e` (the known flake), green on rerun; `sigchld_blocked_e2e` passed 5/5 here on the merged tree. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `629c4e95`.

PR #382, opened 2026-09-17
kernel: start kernels from a binary that still exists when current_exe() was replaced under the daemon (JB-6)

unarbos:
## The incident

arboslife, 2026-09-16 23:07 and 2026-09-17 01:43 UTC (mobile loop, JB-6; `internal/features-inbox/2026-09-17-mobile-spawn-refused-daemon-binary-gone.md`): every `spawn host:arboslife` refused in 35 ms with `start arbos-kernel serve: No such file or directory`. Root then did the challenge itself, so the worktree path #357 fixed was never exercised.

## Cause

`worker.rs:285` started the worker's kernel from `std::env::current_exe()`. On Linux that is `/proc/self/exe`, which names the file the daemon was started from; after an update replaced the binary by unlink-and-write, the link read `… (deleted)` and `Command::new` failed with ENOENT. The new build was on disk at the very path the daemon had been started from. A daemon that outlives its own binary could not start a single worker, and said only ENOENT.

## Fix

New `binary::kernel_binary()`, the one answer for "which file do I start a kernel from":

1. this process's own file, when it still exists (not `(deleted)`);
2. else the path the process was started with — `argv[0]` resolved against the working directory or PATH at start and kept in memory (`binary::remember_start()`, first thing in `main`);
3. else `arbos-kernel` on PATH;
4. else a refusal: *"this daemon's own binary was replaced or moved under it (<path>), and no arbos-kernel was found at the path it was started with or on PATH; restart the daemon from the new binary"*.

Cases 2 and 3 carry a note — which path was used, and *"restart the daemon to run that build yourself"* — printed on the worker daemon's stderr (and `run`'s), because a daemon running an older build than the file it starts kernels from is worth knowing about.

Used at the three `current_exe()` sites: the worker daemon's kernel start, `run`'s kernel start (`cli.rs`), and the remote install's source binary (`remote.rs`). `arbos-kernel binary` prints the choice and the note, for a person on the box.

## QA

- `binary_replaced_e2e`, on the real mechanism: a copy of the kernel is started (`binary --wait-for <flag>`), its file is **unlinked and rewritten while it runs**, then it answers — the new build at its start path, with the note. **Control**: the same copy, not replaced, names its own file with no note.
- Unit tests: the choice order with a `(deleted)` path, a moved file, PATH, and nothing anywhere; `argv[0]` as a relative path, an absolute path, a bare name on PATH, and a name not found.
[...]

PR #385, opened 2026-09-17
kernel/hub: a process says when its own binary is gone (`binary_gone`); the hub keeps one build per registrant

unarbos:
## Two lies the mesh sweep found

`internal/features-inbox/2026-09-17-running-binary-gone-in-hello.md`, from the mesh worker. Together they are why nobody noticed a dead node for two and a half days.

1. **A process whose binary was replaced under it looks healthy.** It answers `hello`, registers, reports the `git_sha` it was built with — and runs an image no file holds. The sweep found seven such processes on two machines, one since Sep 13; a worker daemon in that state refused every spawn for 5.5 hours (JB-6, #382 made it fall back). Nothing any of them sent said so.
2. **The hub kept one build per machine**, taken from whichever registrant spoke last. On arboslife the `demo` kernel (`b6e7098`) and the worker daemon (`bfb36e98`, image deleted) shared a row, so `/list` said the daemon was on `b6e7098`. The phone loop read exactly that and believed the node current. A roster that reports a build it does not know is worse than one that reports nothing, because people act on it.

## What changes

**`arbos_core::binary_gone()`** — computed live at every send, never cached: the state changes while the process runs. Not a path check: on macOS `current_exe()` is the start *path*, and after an update (stage, rename over) that path exists and holds the new file, so a path check would say "not gone" at the one moment it matters. `binary_identity` takes the file's device, inode, size and mtime at start (`main` first thing) and `gone()` is: the path missing, or `(deleted)` (Linux names the unlinked inode), or the file there now having a different identity. Same answer on both platforms.

Carried on:
- `hello` (`binary_gone: true`, only when true — an old client never sees a new key);
- `/healthz` (`"binary_gone": true|false`, for a tunnel or a script);
- `HubFrame::Register` (only when true), which the worker daemon sends through the same `hub_link::register`;
- `arbos-kernel update --place`: the `serving` line reads `/proc/<pid>/exe` and prints `← binary replaced under it; restart to run <sha on disk>`. This alone would have found all seven.

**`MachineInfo.builds: [RegistrantBuild]`** — one entry per registered process (`role: worker`, or `role: kernel` + `project`) with `version`, `git_sha`, `built_at`, `binary_gone`. The hub keeps them keyed by registrant and drops one when it leaves. Top-level `git_sha`/`built_at` are now **the machine's only when every registrant agrees**, empty otherwise; top-level `binary_gone` when any says so. A client reading only the old fields sees nothing rather than a wrong sha; `builds` has the truth.

Self-restart when `binary_gone` is true stays with the self-update work, as routed; this PR only says it.

## QA

- `binary_gone_e2e`: a kernel served from a **copy** of the binary; before, `hello` has no key and `/healthz` says `false`; a new build is **staged and renamed over the same path while it serves** (the installer's shape; the path still exists); a fresh attach reads `binary_gone: true` on `hello` and `/healthz`. Control: an unreplaced copy — no key, `false`.
- `binary_identity` unit test: a file renamed over the same path reads replaced while the path exists; unchanged does not; moved does; with no recorded start identity only absence counts.
- The CI red on the first push was ETXTBSY at the unit test's `spawn` — another test's fork holding the fresh copy's write descriptor — not the detection; the spawn retries on that errno. Six full lib-suite runs green.
- Hub unit test `a_machine_is_not_one_process_builds_are_per_registrant`: a kernel on `b6e7098` and a worker on `bfb36e98` with its image gone → no machine-wide `git_sha`, `built_at` (agreed) kept, `binary_gone: true`, both builds listed with their own; the worker re-registering on `b6e7098` restores the machine's sha and the key is absent when false.
- `update_cmd` unit test (Linux): a spawned copy of `/bin/sleep` reads `binary_gone: Some(false)`, then `Some(true)` once a new file is renamed over its path. (`update --place` reads `/proc` and says nothing extra on a Mac; `hello`/`/healthz`/`register` carry the identity-based answer there.)
- Green: hub 10/10, core 78/78, `hello_built_at`, `hello_identity`, `port_health`, `feedback_bundle`, `tracing`; desktop `cargo check` (it matches `Hello { .. }`).

## Wire

Additive: `binary_gone` on `hello`/`register` only when true; `builds` and `binary_gone` on `MachineInfo` only when non-empty/true. The one semantic change: a machine row's top-level `git_sha`/`built_at` go empty when its registrants disagree — the fix for lie 2. Clients that want a per-process answer read `builds`.


unarbos:
Merge steward: `main` is now `10e11f1d` — #394, #400, #401, #399, #397, #395, #404, #396, #402 and #385 merged; `rust` fast-forwarded to match. #385 conflicted with #400 in `hub.rs` (two adjacent unit tests); both kept, hub suite 11 pass. **Mesh worker:** #394 (log stamps), #400 (name collision refused), #401 (daemon reaps kernels) and #385 (per-registrant builds, `binary_gone`) all want the ArbosLife hub and daemon on this head. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `10e11f1d`. #403 (stacked here) needs retargeting to `main`.

unarbos:
Merge steward, for the record: `update_cmd::tests::a_process_whose_file_was_unlinked_reads_binary_gone` is flaky under CI load — it failed on three unrelated PRs this hour (#405, #410, #403) at line 409, `pid_binary_gone(pid) == Some(false)` reading `Some(true)` right after spawn. Shape: the child has forked but not yet exec'd the copied `sleep`, so `/proc/<pid>/exe` still names the test binary (or a transitional image) and reads as gone. Passes 4/4 alone and 2/2 in the full lib suite here; `main`'s own CI has passed so far, but this will hit a `main` run and skip a dev publish eventually. A poll until `/proc/<pid>/exe` resolves to the copy before the first assertion would close it.

PR #384, opened 2026-09-17
Cycle 28: a settled fold is shut, the newest turn too (F-127); the rig audit's R1/R5/R10 in the gate

unarbos:
## Cycle 28 (stacked on #381 until it merges; retarget to `main` then)

**Gate:** `ui_pass.py` on this build (desktop `5531bd3d`, kernel from `main` `d73a25aea876` — read from the binary, not assumed): **107 pass · 1 fail · 20 not-reachable · 34 unverified · 7 skipped**. The fail is `rewind-turn` "Rewind here (idle) → no state change", once; it passed three times on cycles 25 and 27 after F-122's fix, so F-122 is reopened as *recurred once* rather than closed. `composer-context` reads `unverified` (a dead control or label); `work` reads `not-reachable` — see R10 below.

### App
- **F-127 — a settled Worked fold is shut, the newest turn too.** Cursor's settled headline stands over its timeline in both chat styles ("Worked 1m 15s", `cycle-23/cursor-13`; the worker tab in `cycle-22/`); ours had kept the newest turn open since cycle 3, which read as a timeline the person had asked for. `auto_work_open` is now `running` alone; the unit test's expectation follows. Seen on d11's settled turn (`cycle-28/arbos-d11-settled-before-f127.png` is the before).

### Rig (from `internal/rig-audit-where-a-pass-can-lie.md`)
- **R1** — `seen(id)` (exists *and* `find().visible`) for rows that mean "the person sees it"; used on the kickoff view and the PRs pill to start; the sweep continues.
- **R5** — after a `recover` moved the run to a fresh chat, later rows in that scenario record `pass-after-recover`, not `pass`.
- **R10** — a turn of the root's own (`P_OWN`: run `ls` yourself) before the fold row, so the fold under test is the turn under test. Not enough yet: the row still finds no clickable fold on screen after it — the root's own run under a bare headline draws its rows without a fold line. Open.
- **R11** (new, from the mesh sweep) — a kernel serving a deleted binary. The ArbosLife parity kernel ran 2 d 15 h on a binary deleted Sep 14 13:10; its log shows no attach after the deletion, so no cycle measured against it. Restarted on the on-disk build `52cb63a04075`; rule added: read `/proc/<pid>/exe`, `--version` and `kernel.json.git_sha` before any remote scenario and put the build line in the cycle's note.

### Pairing note — what this rig cannot see
[...]

PR #390, opened 2026-09-17
rewind: a checkpoint needs no git identity; one that could not be saved says so; a restore never cleans on a tree it does not know (qal-j08)

unarbos:
## The data loss

`internal/qa/bugs/qal-j08-…`, measured by QA at `main` `7f6a6b9a` and `1a0003a2`, `rw-04` failing 2/2 on both, `rw-01` (identity set) passing as control. In a repository with no `user.name`/`user.email` — a new machine, a fresh place, exactly J1's user — `commit-tree` failed for want of an identity, `work_commit` returned `None` **silently**, every checkpoint carried HEAD alone, and a rewind with `files: true` ran `reset --hard` + `clean -fd`: every untracked file in the project gone, the kept turns' included, while the `rewound` frame reported success. Rewind is what Jacob reaches for when something has gone wrong, in the state a new project is in.

## Three things were wrong; all three answered

**An internal ref needs no real identity, so the dependency is gone.** The checkpoint commit is written with `GIT_AUTHOR_*`/`GIT_COMMITTER_*` set to `arbos` / `arbos@kernel`. It is the kernel's own ref under `refs/arbos/cp/`, never on a branch, never pushed. Environment beats config, so a bad global config cannot break it either.

**A checkpoint that cannot be written is never silently empty.** `work_commit` now returns `Ok(Some(commit))`, `Ok(None)` (tree equals HEAD's), or `Err(why)` with git's own words. The checkpoint records `clean: true` for the second and `work_error: "<why>"` for the third, and says so on stderr. `work: None` alone no longer means two things. `Checkpoint::knows_tree()` is the one question a restore asks.

**A restore never `clean`s on a checkpoint it knows has no tree.** `restore` refuses `files: true` on a checkpoint with `work_error`, or on a line from before this record (`work: None`, no `clean`): no `reset --hard`, no `clean`, an error naming the reason — *"no checkpoint of the working tree for this turn (…); files left as they are — the transcript is rewound"* — through the existing `Frame::Error` "transcript cut, files not restored" path. A `clean: true` checkpoint may reset and clean, because HEAD alone is the whole truth of that moment.

**`undo` had the same hole**, model-invoked: its mark held HEAD alone and it `clean`ed every untracked file in the project, the user's pre-existing ones included. The mark now carries a second line — the work commit, `clean`, or `error:<why>` — written by the same `snapshot_turn`. Without a known tree, `undo` resets tracked files and says *"untracked files left as they are (no checkpoint of the working tree: …)"*; with one, it brings the turn-start tree back whole (`read-tree`), which it never did before.

## QA

Engine unit tests on an **identity-less repository**: the repo's own config sets `user.name`/`user.email` to empty (which beats this box's global identity) plus `useConfigOnly`; the fixture first proves git itself refuses `commit-tree` there, or the test means nothing.
- `a_checkpoint_needs_no_git_identity_and_rewind_brings_the_files_back` — the work commit is made; `snapshot_turn` records it and the `undo` mark's second line matches; a restore brings back the kept file and removes the later one.
- `a_checkpoint_that_could_not_save_the_tree_says_so_and_restore_will_not_clean_on_it` — an old-shape line and a `work_error` line both refuse with the reason and remove nothing; a `clean: true` line restores.
- `undo_without_a_known_tree_resets_tracked_files_and_leaves_untracked_ones` — a HEAD-only mark: tracked file reset, the user's untracked file kept, the result says why; a `clean` mark: the turn's file removed.

`rewind_no_identity_e2e` — QA's `rw-04` through the kernel:
- `rewind_with_files_in_a_repo_without_git_identity_keeps_the_kept_turns_files` — three turns (`echo first > f1.txt`, `echo second > f2.txt`, a third), no identity; checkpoint 1 `clean: true`, 2 and 3 carry `work`, none `work_error`; rewind to turn 3 with files → `restored` names the working tree, `f1.txt` and `f2.txt` stand with their contents.
- `a_rewind_on_a_checkpoint_without_a_tree_leaves_the_files_and_says_so` — checkpoints rewritten as an old kernel wrote them (HEAD alone); the rewind answers `error: rewind: transcript cut, files not restored: no checkpoint of the working tree …` and `f1.txt` stands.

**Control at `main` `0f2a8bc6`** (worktree): both fail — checkpoint 2 has no `work`/`clean` (`{"head","line","ts"}` only), and the old kernel reports `restored: 65eb841c6397` on a HEAD-only checkpoint, the exact frame QA saw.

Regression green: `standing_pass_e2e` (3/3, including the rewind pass), `git_author_e2e`, `fix_on_branch_e2e`, `worktree_cleanup_e2e`; engine lib 83/83.

## Compatibility

`checkpoints.jsonl` gains two optional keys; old lines read as "tree unknown" and are refused for `files: true` rather than trusted — the safe reading. The `undo` mark gains a second line; a one-line mark reads as "tree unknown". `rewind` without `files` is untouched.


unarbos:
Merge steward: merged as 011a6cf0 on QA's word (`qal-j08` closed at `9ade320f`, `qal-j11` for undo passing); `main` is now `011a6cf0` — also carries #388, #386, #384, #391, #387; `rust` fast-forwarded to match. Destructive paths re-read at the head: `restore` bails before any `reset --hard`/`clean` unless the checkpoint knows its tree; `undo` on a one-line mark resets tracked files only. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `011a6cf0` — this is the data-loss fix Jacob is waiting on.

PR #392, opened 2026-09-17
record: absent is not unknown, and nothing destructive acts on unknown (qal-j09, j10, j12, j13 — the class behind qal-j08)

unarbos:
Stacked on #390 (it builds on the checkpoint record #390 introduced). Base: `cursor/rewind-no-identity-b027`; retarget to `main` once #390 lands.

## The class

QA's census of the qal-j08 family (`docs/qa-loop-design.md`, "The qal-j08 family"): 34 best-effort writes and 12 `unwrap_or_default` reads. The common fault: a read helper that cannot tell **absent** from **unknown**. `read_to_string(..).unwrap_or_default()` turns EIO, EACCES, a lock, a partial view on a mount, into `""`, and the next writer acts on that emptiness with confidence — the project page rewritten to one line, a checkpoint carrying HEAD alone, `undo` to a stale mark. A file that is not there is a fact (start empty). A file that could not be read is a question, and a question is not a value.

## The helper — the deliverable

`arbos_core::record`:
- `Read<T> = Present(T) | Absent | Unknown(why)`. `read_text`, `read_json`.
- `.confirmed() -> Result<Option<T>>`: the record, `None` when confirmed absent, an **error** when unknown — never a default. `.present()` for display-only readers, named so the grep finds anyone who displays and then writes.
- `write_atomic(path, bytes)`: temp + rename, whole or not at all, the path in the error.

**The rule, made explicit** as design goal 12 in `docs/design/filesystem-state-design.md`: *Absent is not unknown, and nothing destructive acts on unknown.* No `git reset`, no `clean`, no file rewritten over its old bytes, on a record the kernel has not confirmed. Reads that feed a write go through `.confirmed()`; records later readers act on are written with `write_atomic`; a failed write is said rather than swallowed. `grep unwrap_or_default` near a `write` is the smell.

## The two driven bugs, inside the change

**qal-j09 — one bad read wipes the coordinator's page.** `Notes` now carries whether its read failed. Such a page is `unread`: it displays as empty (so nothing that shows a page changes) and **`save_path` refuses to write it** over the file, with the reason. The `plan` call fails "could not read .arbos/notes.md: … — nothing written; the page keeps its bytes", instead of "Set 1 item(s)." This protects every writer of the page at once, including ones nobody has found yet.

**qal-j10 — a failed turn-start mark makes `undo` destroy a commit.** The mark is now written atomically and carries `line:<n>`, the turn it is for. The old mark is removed *before* the new one is written, so a failed write leaves **no** mark, never an older turn's; the failure lands on the transcript as a failed notice ("Checkpoint not written for this turn: … undo and a rewind of files to this turn are refused rather than guessed"). `undo` reads the mark through `.confirmed()`, and refuses one for another turn, one that names no turn, or none at all: "no checkpoint for this turn …; nothing reset". `RunCx.turn_line` carries the turn to the tool; the kernel's own `Frame::Undo` passes the last checkpoint's line and now reports a refusal instead of `let _`.

## The rest of the destructive census

- `inflight::start` — the qal-j02 record was documented best-effort; a full disk silently returned the run-twice hole. It now returns an error and **the tool call is refused** with the reason rather than run unrecorded.
- `memory` (`remember` add, `forget` index, the entry index), `hooks::notify_user` (`user.md`), `files::init_arbos_repo` (`.arbos/.gitignore`), both `.git/info/exclude` writers, `plan.rs` `meta.toml`, `host::remember_place`, `repro.jsonl`: confirmed reads, or plain appends where a read was never needed. None rewrites from a failed read.
- Left as misreport-only, listed for the next pass: `jobs.rs` `killed`/`settled`/`detached` markers, `blocked.rs` rename, `mechanism.rs`, `provider.rs`, `protocol.rs` — none of them precedes a delete or a reset.

## qal-j12 and qal-j13, folded in (`92f6eb59`, `3fcae4f1`) — so Jacob's update is safe in one press

The one-time migration wrote its new records and then `let _ = rename(plan.jsonl → .migrated)` — the only record that it had happened. When that rename failed once (a read-only agent folder with writable subfolders, a permissions slip, a partial view on a mount), the next start migrated again: the standing cron existed twice and fired twice every period for ever; the pending task was queued twice. It triggers on the first start after an upgrade of every place that still has a `plan.jsonl` — Jacob's older projects, on the update he is about to press.

It is the rule arriving one step late, so the rule is applied to its own step: **the completion record comes first.** The old file is renamed to `plan.jsonl.migrating` *before* a single new record is written.
- Cannot be moved aside → **nothing migrated**; `migrate_blocked` in the log; a failed notice on the transcript: *"The one-time migration of the old plan (plan.jsonl) could not start: … Nothing was migrated and the old plan stays as it is; fix what blocks writes in this agent's folder and start the kernel again."*
- Moved → migrated from the moved file, then renamed `plan.jsonl.migrated` (a failure there is logged and harmless).
- A `.migrating` file with no `plan.jsonl` — a start cut between the move and the finish — is **finished, not left** (qal-j13, `3fcae4f1`): the records the cut start already wrote are recognised and not written again (a subscription with the same kind, period and command — or prompt —, an inbox file with the same sender and body, a parked question with the same id, a checklist line with the same text), the rest is carried over, and the source becomes `plan.jsonl.migrated`. The transcript says both counts: *"An earlier start began carrying over the old plan (plan.jsonl) and was cut before it finished. Finished now: 1 pending task(s) carried over this time; 1 were already in place and were not written again."* The first version of this fix kept the file but only the log said so — a standing cron the person had was simply gone from what runs; the same lesson one turn later.
- A completed first-time migration also says on the transcript what it carried over and where the old file is kept.
- Same for `subscriptions.json`; the two harmless moves (`attempts.jsonl`, `plan.md`) log their failures instead of `let _`.

QA: unit test `a_migration_that_cannot_record_itself_writes_nothing_and_never_doubles` (read-only folder → nothing, twice, and the notice; writable → one cron and one task after two runs; a cut with the cron written and the task not → the cron recognised, the task carried over, both counts said). `confirmed_reads_e2e::a_cut_migration_is_finished_without_doubling_and_the_transcript_says_so`: the cut shape on disk through the kernel — the missed task runs at boot, exactly one cron, the notice with both counts; **control at `92f6eb59`**: the missed task never runs. `confirmed_reads_e2e::a_failed_migration_record_writes_nothing_and_a_completed_one_is_not_repeated` is QA's `sw-05` through the kernel: two starts on the read-only folder → `plan.jsonl` stands, zero crons, two `migrate_blocked` lines; two starts writable → exactly one cron. **Control at `a0f2a92d`** (this PR before the fix): two crons after two starts — QA's doubling.

On QA's misreport note: agreed, no action — an unwritten `detached` marker means the job ends with its kernel, which is the leash's default and the safe side.

## The Jacob-facing refusal

#390's "no checkpoint of the working tree" arrived as an `error` frame with no transcript notice, so a window drew a rewind that had half-worked as a failure (the qal-j05 shape), and on his existing places most old-turn rewinds land there. It is now a typed `NoTree`; the kernel writes a plain `notice` — *"Rewound the transcript. Files were not restored: … This turn was recorded before the kernel kept each turn's working tree; turns recorded from now on restore their files."* — and closes the pending `rewound`. Real git failures still arrive as errors.

## QA

`confirmed_reads_e2e`, QA's scenarios through the kernel with QA's injectors:
- `an_unreadable_project_page_is_not_rewritten_by_plan` (`sw-01`): a 293-byte page `chmod 000` for the call; afterwards the bytes are identical and the `plan` record carries "could not read … not written".
- `a_stale_undo_mark_is_refused_and_a_kept_commit_stands` (`sw-02`): turn 1 commits A; the mark and `runtime/` made read-only; turn 2 commits B and calls `undo` → refused "no checkpoint for this turn … nothing reset", commit A and `a.txt` stand, HEAD is B, exactly one "Checkpoint not written" notice.
- `a_failed_migration_record_writes_nothing_and_a_completed_one_is_not_repeated` (`sw-05`) and `a_cut_migration_is_finished_without_doubling_and_the_transcript_says_so` (qal-j13), above.

**Control at #390's head `9ade320f`** (worktree): `sw-01`/`sw-02` fail as QA measured — the page becomes `## Checklist\n- [ ] Add perimeter() tests — ready\n`; `git log` shows only `start`, commit A gone.

`rewind_no_identity_e2e`'s refusal case now asserts the notice (failed: false, "turns recorded from now on restore their files") and the closing `rewound`, not an `error`. Engine unit tests for `undo` updated to the two-argument form; `record` has its own (absent vs unknown, unparseable is unknown, atomic write). Green: core 77, engine 83, kernel lib 40; e2e `standing_pass`, `page_algorithm`, `page_template`, `focus`, `correction_nudge`, `archive_children`, `archived_children`, `restart_mid_turn`, `restart_states`, `kickoff_turn`, `subscriptions`, `github_prs`, `worktree_cleanup`, `git_author`, `done_wake_fold`; desktop `cargo check`.


unarbos:
Merge steward: merged on the coordinator's call — QA's verification at `a0f2a92d` plus the author's conflict resolution at `ba4333de`; the four later commits (migrate completion-first, the deferred markers, `inflight::start` checked) add refusals and `.migrated` renames, no new deletion. `main` is now `f36c660d`, also carrying #444; `rust` fast-forwarded to match. #444 overlapped this PR in two places and I composed them: `exclude_locally` now creates the directory with context (#444), reads confirmed (#392) and returns an unknown read as the error the caller says on the transcript; the undo mark is written atomically with the `line:` field (#392) and removed on a failed write so a stale mark is never used (#444). Workspace 793 pass. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `f36c660d`. #398 (stacked here) is superseded by #405 — closing it separately.

PR #391, opened 2026-09-17
Cycle 29: the kernel's own --version as every run's first row; a refused rewind says why (F-122 traced)

unarbos:
## Cycle 29 (stacked on #384 until it merges; retarget to `main` then)

**Gate:** `ui_pass.py` — first row now the kernel's own word: `arbos-kernel 0.2.0 d73a25aea876 protocol 1`. Full run on `d27eb13a`: **107 pass · 1 fail · 20 not-reachable · 34 unverified · 7 skipped · 1 info**, the fail `rewind-turn` — which this build then traced and fixed (below); turn phase on the final build `6f7b757b`, twice: **33 · 0** and **34 · 0**, `rewind-turn` passing both. `composer-context` reads `unverified` (dead control or label).

### F-122b — traced from the app's own line, then fixed
A refused rewind now logs its state. On this run: `streaming=false turn_open=true working=false last_frame=0s progress=0s turn_ended=None` — the turn had ended (the gate saw it idle), then the kernel's settled records landed **past the 1 s tail lag** and `turn_alive` reopened it, clearing `turn_ended`. Fix: a record arriving after the turn's own end never reopens it, however late; a new turn announces itself first — a prompt sent here, a wake, another client's line, or the kernel's thinking pulse — each clears `turn_ended`, and only then does a token reopen the turn.

### Rig
- **R11** — every gate and journey run records `arbos-kernel --version` from the binary it launches as its first row (`kernel | rig | info`), never from a roster or the branch. Prompted by the ArbosLife kernel that served a deleted binary for 2 d 15 h; read from its place: no cycle measured against it.
- **R12** — a claim about another component's behaviour quotes the lines it rests on or is marked *inferred*. Earned twice tonight: my `unknown`-sha reading of the stranger control was backwards (inferred, not traced; #388 fixes the real gap), and #372's merge state was misread from its PR page (the tree says it is in `main`).

### Not done this cycle, and why
The Cursor side of d11 — Cursor's renderer would not launch on the rig (`launch-failed 1002`, `GpuControl.CreateCommandBuffer` transient failures) after a hard kill earlier tonight; its processes are stopped and a clean relaunch opens cycle 30. The ArbosLife reinstall through the app's own path and light/dark side by side carry too.

Kernel CI, if red: `standing_pass_e2e` is the known flake; this diff is `desktop/` only.

[...]

PR #395, opened 2026-09-17
Cycle 31: the deliberate pass over the gate's assertions — eight could never fail (R14); seen() and the window read back

unarbos:
## Cycle 31 (stacked on #391 until it merges; retarget to `main` then). Rig only — no app code.

**Gate:** `ui_pass.py` on the cycle-29 build, kernel `arbos-kernel 0.2.0 d73a25aea876 protocol 1` (read from the binary): **108 pass · 1 fail · 19 not-reachable · 35 unverified · 7 skipped · 1 info**; the fail (`pill-prs`) was R15 below, and the PRs phase rerun after the fix passes (6 · 0). `composer-context` reads `unverified` (dead control or label). The `work` fold row **passes** for the first time since cycle 23: R10's own-turn prompt gives it a real fold, and it measures rows.

### The deliberate pass — `internal/gate-assertions-what-each-reads.md`
All 119 `check`s classified from the AST: 63 read a state field, 20 existence, 18 derived, 17 assert nothing (already `unverified`), 1 geometry. Found and fixed the same hour:
- **R14 — assertions that could not fail.** `self.diff(a, b) or "clicked"` (×4) and `… or "card still up"` returned a truthy string on no change; three bare f-strings (`f"{n} rows…"`) were truthy at n=0. Eight rows that had never been able to fail now return `unverified: …` and record so. Rule: an `ok` returns `True`, `False`, or a string starting `unverified:`.
- **R1 applied** — positive `exists("X")` where the row means "the person sees X" (model list, commands list, stop disc, tab sheet, mic test, composer field, Working card, follow-ups head) reads `seen("X")`. Negative existence stays.
- **R15** — the driver's `visible` (bounds ∩ content mask) read `False` for the PRs pill while it was plainly on screen (`128-pill-prs.png`); `seen()` takes `reachable` (the hit test at the centre) as well.
- **R13, raised and withdrawn.** "The window a third off-screen" was my crop of a still (cut at x=200; the gate places the window at x=100). The raw still shows it whole; cycle 29's results stand. The geometry read-back stays as a guard — and it already paid: the window the gate asks for (1600×1000 at 100,60) is granted as **1100×761 at 102,116**, so every gate still is of that size; now logged per run.

Kernel CI, if red: `standing_pass_e2e` is the known flake; this diff is `desktop/qa` only.


unarbos:
Merged into `main` as 0a375947. Closed rather than shown as merged because this PR's base was a stacked branch; head `bed04cd` is fully contained in `main`.

PR #398, opened 2026-09-17
rewind: the checkpoint's record lands before the turn goes on; the tree follows and the store never sinks it (kills the standing_pass_e2e flake — a product race)

unarbos:
Stacked on #392 (one level; it builds on the checkpoint record there). Retarget to `main` once #392 lands.

## The flake was a bug

`standing_pass_e2e::rewound_arrives_before_the_file_restore` has been red under CI load for days and green on rerun, and it cost two gate judgement calls tonight. QA's 32 sampled rewinds showed the transcript on disk is never emptied or short, which was right — the file was never *corrupted*. It was **cut in the wrong place**.

`snapshot_turn` ran whole on the blocking pool, un-awaited, at turn start. `rewind turn N` resolves to the checkpoint on or before the Nth user line. A rewind that arrived before turn N's checkpoint had landed resolved to turn N−1's checkpoint and cut one turn too many — down to an empty transcript when N−1 was the first turn, which is the `[]` the test saw. Driven under starvation (test pinned to one core beside four spinners) it failed 1 in 25, then 7 in 20 once the second fault below was exposed. In life: a person who presses Rewind right after a turn on a busy machine loses a turn from the record.

## Two fixes

**The record before the turn goes on.** `snapshot_turn_record` (cheap — `rev-parse HEAD`, the line, the undo mark cleared and set) is awaited before the turn continues, its tree marked pending (`work_error: TREE_PENDING`). `snapshot_turn_tree` (the `git add -A` into a scratch index, the ref, the mark with the tree) follows on the blocking pool, writes the filled record to `checkpoints.d/<line>.json` and rewrites the line in `checkpoints.jsonl`. So a rewind at any moment resolves to the right turn. `restore_files` waits up to 20 s for a pending tree through the sidecar before it refuses with the reason — a rewind pressed right after a turn restores its files rather than saying "still being saved".

**The store cannot sink the tree save.** The same starvation exposed a second fault: every tree save failing with *"'.arbos/' does not have a commit checked out; fatal: adding files failed"*. `.arbos/` is a git repository of its own (the F design's Phase 1), and `git add -A` refuses an embedded repository with no commit when the project does not ignore it. The kernel writes `.git/info/exclude` at start — but a **fresh place whose agent runs `git init` in turn one** has no exclude until the next kernel start, so every checkpoint in that first session had no tree, and `rewind --files` refused (safely, since #390; silently and destructively before it). A pathspec exclusion (`:!.arbos`) fails the *other* way when `.arbos` *is* ignored ("paths are ignored by one of your .gitignore files"). `work_commit` now adds with an excludes file of its own — the user's global excludes plus `/.arbos/` — which works in both states.

## The test

Reads the last *turn* event (a kernel notice after `turn_complete`, saying what the restore did, is not the turn); accepts the closing `rewound` with or without a restore (that is "the restore reports"). The empty-transcript assertion stands as written and is now the guard for the first fix.

## Numbers, at named commits

- **This commit `af9eafe3`, starved** (one core, four spinners): **0/25** failures. Idle: 3/3.
- **Control `60ce1405`** (the base), same starvation, same test file: **1/15**; with the old test file, **7/20** (the empty-transcript failure, then the tree-save failure once the first fix exposed it).
- Regression green: `rewind_no_identity_e2e` (3), `confirmed_reads_e2e` (4), `restart_mid_turn`, `restart_states`, `git_author`, `fix_on_branch`, `worktree_cleanup`, `kickoff_turn`; engine `tools::git` 12/12 including the new `the_checkpoint_record_is_on_disk_before_its_tree_and_is_filled_in_after` (pending record refused for files with "still being saved"; filled in place, not appended; the undo mark carries the line only once the tree lands).

## Cost

Turn start awaits one `git rev-parse HEAD` and an append (~milliseconds) before the model is called; the expensive add stays off the turn's path. One extra small file per turn under `checkpoints.d/`, removed after a restore uses it.


unarbos:
Superseded: the same fix landed as #405 (rebased onto `main`, plus the follow-up that settled the `update_cmd` test); `git cherry` finds nothing here that `main` lacks. #392, its base, is merged in `f36c660d`.

unarbos:
Correction to the note above: `git cherry` does mark `af9eafe3` as not on `main` — its patch differs textually from #405's `94f4eb9d` because #405 was re-done against `main` (the author's own description: "the same fix as #398, rebased onto main"), not because content is missing. The closing stands on that description and on #405 + its follow-ups being merged, not on a clean cherry.

PR #403, opened 2026-09-17
kernel: a kernel whose file was replaced under it restarts onto the new one when idle (no supervisor needed)

unarbos:
Stacked on #385 (one level; it needs the identity-based `binary_gone`). Retarget to `main` when #385 lands.

## Why

Jacob's `subnet120` kernel is started detached with `init` as its parent. When its binary is replaced under it — an update; our own loops installing into the shared `PATH`, since banned — the replacement is permanent for it: it serves stale code until a person notices, where a process under a `while true` loop simply relaunches onto the new file. A kernel with no supervisor has no way to recover.

## What

With `binary_gone` known (#385), the serve loop's five-second tick:
1. says `binary_gone` once in the log when its file is gone or replaced;
2. at the update gate's **Idle** — `idle::update_verdict_quiet`: no turn running, no question waiting on a human, no subscription run in flight, no remote child mid-turn, nothing due within a minute — **execs onto the file at its own path** with the same arguments and environment. Same pid; the place lock is an `flock` on a CLOEXEC fd, released at exec and taken again by the new image; the listening socket is rebound and clients reconnect the way they do after any kernel restart. A notice lands on root's transcript first (*"The kernel's program file was replaced under it (an update); restarting onto the new build now, nothing in flight. Windows reconnect on their own."*) and a plain frame goes to attached clients.
3. An exec that fails **returns** and the old image serves on, logged `reexec_failed` — the property the self-update design chose `execv` for. Tried at most once a minute.

`ARBOS_NO_REEXEC=1` keeps a kernel on its old image: the two report tests use it so they measure the report rather than the restart, and a person who wants to choose the moment can set it.

The same rule for a kernel under a supervisor loop or the desktop: exec, not exit. Exiting would be right only under a loop, and "is there a loop above me" is not knowable in general (the desktop and the worker daemon are parents that do not relaunch); exec is right everywhere.

## QA

`binary_gone_e2e::a_kernel_whose_binary_was_replaced_restarts_onto_the_new_one_when_idle`: a kernel served from a copy; the same build staged and renamed over the copy's path; within two ticks `kernel.json` is rewritten by the new image with the **same pid and a later `started`**; the log has `binary_gone`, `reexec`, and two `kernel_start` lines from one process; a fresh attach's `hello` no longer says `binary_gone`; the transcript carries the notice. The same build is used so the mechanism is proven with a kernel the test can still drive; a different build would exec the same way.

Green: `update_gate_e2e` (the execv survival test from #353 — the two mechanisms coexist), `leash_e2e`, `hello_built_at_e2e`, `serve_e2e`, `port_health_e2e`, and #385's two report tests under `ARBOS_NO_REEXEC`.

## Not done here

A worker daemon (`arbos-kernel worker`) with its binary replaced still needs a restart to run the new build itself — #382 lets it start kernels on the new build meanwhile. The same exec-when-idle could apply to it; its "idle" is "no claim in flight", which the daemon knows. A follow-up if wanted.


unarbos:
Merge steward: green and read, **held on sequencing**. `reexec_onto_new_binary` execs onto `kernel_binary()` with no probe of the new file first, so a published build that starts and dies would take the serving kernel down with it and, under the ArbosLife daemon, leave nothing to respawn it. The pre-swap probe and rollback copy (`63666ae0`, orphaned after #318, re-landing as its own PR from `cursor/reland-the-update-probe-and-rollback-1b59`) is the net for exactly that; this merges right after it. Cheap addition worth considering here regardless: run `<chosen> --version` before `exec` and back off on a non-zero exit, so the re-exec has its own probe.

unarbos:
Merge steward, one suggestion for the author's call, separate from the sequencing hold: `reexec_onto_new_binary` could run `<chosen> --version` (the same probe `update_cmd` uses) before the `exec`, and on a non-zero exit log `reexec_failed` and back off rather than exec — so the re-exec carries its own net whatever the updater did. Cheap, and it makes this path safe on its own; the hold lifts when the probe/rollback re-land is on `main` either way.

unarbos:
Merge steward: merged (`20ba294`, after #418), but one of this PR's own tests is flaky on the merged tree and I want it named before it reddens `main`: `binary_gone_e2e::a_kernel_whose_directory_was_renamed_restarts_onto_the_start_path_not_the_backup` failed 2 of 5 runs here with *"the kernel never restarted after its directory was renamed"* at the 30 s deadline, passing the other three. Shape that fits: the check runs on the 5 s tick, and one non-Idle verdict (or one failed attempt) sets `reexec_backoff_until` 60 s out — past the test's 30 s window — so a single early miss reads as never. Either the test waits longer than `REEXEC_RETRY_MS`, or the first retry after a miss comes sooner than a minute. This was my slip too: the push went out before the test result was read. If `main`'s CI goes red on it I will revert the merge rather than let a publish be skipped.

PR #414, opened 2026-09-17
jobs: a kernel that exits takes its jobs with it — a zombie kernel reads as gone, a stopping kernel ends its jobs itself, `run` says what it leaves serving

unarbos:
SWE-bench loop, cycle 14 (`internal/qa/inbox/2026-09-17-swebench-loop-cycle-14.md`): in 4 of 40 rollouts the sweep after the kernel had exited found 4–10 live processes — the job-runner shells and the tests under them — reparented to PID 1. Without the sweep they would have had the network back at grading.

## Why the leash did not fire

The container's PID 1 is `sleep infinity`, which reaps nothing. `arbos-swe-run` stops the kernel with a TERM to its group; the kernel exits; nobody `wait()`s it; it stays a **zombie**. The leash's kernel check was `kill -0 $K` — and `kill -0` on a zombie **succeeds**. So every job's leash read "my kernel is alive" for as long as the container lived. Same defeat as the `pgrep -g` zombie in #377, one level up. Staged in a test with a parent that never waits (python `Popen`, no `wait`): before the fix the job group outlived the zombie kernel; after, it ends within the poll.

## Three changes, one shape

1. **Leash:** `alive()` reads the process state — `/proc/<pid>/stat` field 3, else `ps -o stat=` — and treats `Z` as gone. Same `killed: the kernel exited and the job was ended with it`.
2. **Kernel graceful stop** (TERM/INT — `arbos-kernel stop`, the harness, Ctrl-C): after ending turns it ends every running job itself (`end_jobs_for_stop`), each folder saying `killed: the kernel was stopped and ended its jobs with it`, logged `kernel_stop_jobs`. The leash stays the backstop for a kernel that *dies*; a kernel that is *asked* to stop knows what it started.
3. **`arbos-kernel run`** that started the kernel used to exit saying nothing about it. It now says, on stderr: the kernel serves on (pid), which jobs still run under it (agent, id, command), and how to end them (`arbos-kernel stop <place>` or a TERM to the pid). With `--json`, one line `{"kind":"kernel_left_serving", "pid", "place", "running_jobs":[…]}` on stdout. Nothing is said when the kernel was already there — that one is someone else's. The kernel serving on is what makes the next `run` fast, so `run` does not stop it; it says so.

I chose "end them and say so" for the kernel's own stop, and "say what stays" for `run`, rather than `run` refusing to exit: `run` is a client, the jobs belong to the kernel, and a background job (a dev server) outliving one prompt is the point of `background:true`.

## Tests (leash suite 12/12)

- `a_kernel_left_a_zombie_by_a_parent_that_never_reaps_still_takes_its_jobs_with_it` (linux): the staged fault above; asserts the kernel really is `Z` and `kill -0` still says yes before checking the job ended and the folder says why.
- `a_kernel_asked_to_stop_ends_its_jobs_itself_and_says_so`: INT to the kernel → group gone, folder says "the kernel was stopped".
- `run_that_started_the_kernel_says_what_it_leaves_running`: `run` spawns a replay kernel from env, the turn backgrounds a loop, `run` exits 0 and names the pid, `1 job(s) still running`, `root j1: while :; do echo tick`, and `arbos-kernel stop`; the TERM it names ends the job.

Green: `run_no_prompts`, `stop`, `update_gate`, `bash_background`.

## For the harness

With 2., the harness's step 5 (TERM to the group, wait ≤10 s) should leave `left 0` without the sweep; with 1., even a KILL does. `sweeps.json` on the next cycle is the outside check.


unarbos:
Merge steward, for the record: `job_leash_e2e::a_kernel_asked_to_stop_ends_its_jobs_itself_and_says_so` fails deterministically on `main` **on this box** (4/4 at `663d8ce`, also in a clean worktree of `main`) while passing on your branch tip here and on `main`'s CI runner. The `killed` file reads `killed by the kernel` — `Root::kill`'s own marker — not the stop message: the turn-stop path has already ended the job by the time `end_jobs_for_stop` runs, so `kill` returns `Ok(false)` (not running) and the stop wording is never written. Environment-sensitive (timing of the turn stop vs the job sweep), and the job does end either way; only the wording and the test's expectation differ. Worth a look — either `end_jobs_for_stop` also rewrites the marker for jobs the turn stop just ended, or the assertion accepts both.

PR #418, opened 2026-09-17
update: re-land the pre-swap place probe and the .previous rollback copy

unarbos:
## This is a gap in `main`, not an improvement

`main` has been missing both of these since **14:27 UTC yesterday**, and it is the path Jacob is about to use.

The commit was written at **14:29:24** — two minutes *after* [#318](https://github.com/unarbos/arbos/pull/318) merged at **14:27:24**. I pushed it to a branch whose PR had already closed, so it never landed and nobody noticed. `63666ae0` is cherry-picked here unchanged.

I found it only because later work would not compile against `Probe`. A compiler error, two hours of work later, was the only signal.

## What `main` does today, precisely

Worth stating exactly, because it is less bad than "no probe" and worse than it looks. `main` **does** check the staged binary before swapping:

- it runs `--version`, so a payload for the wrong architecture, or one that will not execute at all, is caught **before** the old binary is touched;
- `Swap` still rolls back from `Drop` if the post-swap check fails.

What it lacks is the two things that matter when a build *runs* and is still wrong:

| missing | consequence |
| --- | --- |
| `Probe::Place` | a build that answers `--version` but cannot read the store is installed anyway |
| `<bin>.previous` | once `swap.commit()` unlinks the old binary, there is **no local way back** |

## So what happens today if a published build is broken?

- **It will not execute** (bad arch, not a binary) → caught before the swap. The old kernel survives. Fine today.
- **It runs `--version` and then dies at boot, or cannot read the store** → passes both of `main`'s checks. The swap commits, the old binary is unlinked, and the machine is left with a kernel that will not start and **no copy of the one that did**. Recovery needs the network — re-fetching an older build from the feed, which keeps the last three — or a rebuild.

That second case gets worse once [#403](https://github.com/unarbos/arbos/pull/403) lands: `execv` succeeds, because the file *does* run; the new image then dies at boot; and on a box with no supervisor — `subnet120` is exactly that — there is no kernel at all and nothing to bring one back.

## What this restores

**`Probe::Place`** — the staged binary must read a real place *no worse than the binary it replaces*, via `arbos-kernel check`, which walks the store with the parsers `serve` boots on. Differential rather than "must be clean", because a place with pre-existing errors is not the new build's fault and refusing on it would block updates on exactly the machines that need one.

Driven here against a stub that answers `--version` correctly and then fails — the shape `--version` cannot catch:

```
REFUSED: the downloaded kernel finds 2 problems in /tmp/.../place
         where the one it would replace finds 0 — refusing it rather than serving with it
still the good build? arbos-kernel 0.1.40 goodbuild001 protocol 1
beside it: arbos-kernel
```

Refused, nothing moved, nothing left behind.

**`<bin>.previous`** — the replaced binary is kept beside the new one. Not hidden, unlike the staging and backup names: somebody looking at a kernel that will not start should find the one that did sitting next to it. It turns "the new kernel dies at boot on a box nobody watches" from unrecoverable into one `mv`, and `arbos-kernel update`'s restart note says where it is.

## Why the probe is load-bearing rather than belt and braces

`execv` reports an error only when **the exec itself** fails. A binary that starts and then dies while booting returns nothing, because by then the old image is gone — there is no "it came up, so keep it" moment afterwards. Every bit of confidence has to be bought *before* the old binary is moved aside, and `<bin>.previous` is the answer to the failure no probe can see.

## How to check

- [x] `cargo test -p arbos-update` — 16 in `kernel::`, including both probe branches and the version mismatch.
- [x] The place probe driven end to end against a stub that passes `--version` and fails `check`.
- [x] `cargo test --workspace --locked` exits 0.

One caveat I would rather state than hide: on the **first** full run one test failed and the summary line was lost in the output. Two subsequent full runs exited 0 and I could not reproduce it, so I believe it is a flake — but I did not identify which test, and I would not want that read as a clean sweep.

## Notes for the reviewer

- This is `63666ae0` cherry-picked, not rewritten. It was reviewed once as part of #318's branch before the merge raced it.
- `install_payload` and `verify_and_install` take a `Probe` argument. `arbos-updatectl` passes `Probe::Version`; `arbos-kernel update` passes `Probe::Place` when it knows a place and `Probe::Version` otherwise.
- Next on this path: `update --install --binary <target> --from <file>`, installing a kernel already on the machine with this same probe and rollback — the primitive the remote bootstrap needs, since `ssh_install_kernel` currently does `mv -f` over the target and a box with no route to GitHub cannot use the feed. Held until this lands rather than built on an unlanded commit.


unarbos:
Merge steward: merged as a fix — `main` is now `20ba2947` with #418, then #421, then #403 in that order; `rust` fast-forwarded to match. On your lost failure: two full workspace runs here on the merged tree were clean (753 pass, 0 fail, every summary line present), so I cannot name it from my side. The one name I do have: your CI's first attempt failed on `restart_states_e2e::a_parent_waiting_on_a_worker_picks_up_after_a_restart_and_hears_the_report_once` (#342's test, untouched by this diff), green on rerun and 6/6 here under load — a candidate, not a proof. **Mesh worker:** `subnet120` and the ArbosLife daemon want this head: it carries the probe/rollback net and the idle re-exec together. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `20ba2947`.

unarbos:
Merge steward, for the record: `kernel::tests::the_place_probe_allows_a_build_that_reads_it_no_worse` is flaky under parallel load — failed once in a full workspace run here (`kernel.rs:738`, the same-count stub) and once on #425's CI; 3/3 green alone. Same shape as the `binary_gone` unit test: a stub script spawned while other tests fork. A retry on ETXTBSY / a settle before the first `run` would close it.

PR #419, opened 2026-09-17
rewind: a restore that fails leaves the tree where it was — objects checked first, the current tree kept and put back on any later failure; a failed clean is in what "restored" says (qal-j16)

unarbos:
Two commits. The first (`0f71cca4`) moved `clean` after `read-tree` and put a failed clean into the restored text; QA verified both claims against a control (`rw-08`, `rw-09`). Their general assertion then caught what the first commit did not reach — **qal-j16**: `reset --hard <cp.head>` was still the first step, so a `read-tree` that failed on a missing object left HEAD moved back and the person's own later commit gone from the working tree, while the message said "files not restored".

## The property

*A restore that reports an error leaves HEAD, the index and the working tree exactly as they were.* That is the acceptance criterion here, not the one bug, and it is the assertion in both new tests (`state()`: HEAD, `ls-files -s`, every file's bytes).

## Two mechanisms (`0bceb0df`)

1. **Check before anything moves.** `git cat-file -e <head>^{commit}` and `<work>^{tree}`. A missing or corrupt object is refused up front with the object named: *"the checkpoint's working tree (8c4fc5ab11d5) is not in the repository (a missing or corrupt object); files left as they are"*. This is QA's staged case.
2. **Keep the tree, put it back.** Before the first destructive step, the tree as it stands is committed the same way a checkpoint's is (`work_commit` on the current HEAD). On any later failure — `reset --hard` or `read-tree` — HEAD and that tree are put back with the same `reset --hard` / `read-tree -u --reset` / `reset -q`, and the error says: *"git read-tree … failed; the tree was put back as it was (HEAD 3e05ca528c28, your files from 9f1e…)"*. If the put-back itself fails, the error carries the exact commands to recover by hand. A tree that cannot be copied refuses the restore. This holds for failures nobody anticipated, which is the point.

## Tests

- `a_restore_whose_checkpoint_is_corrupt_is_refused_with_the_tree_untouched`: the checkpoint's work commit object removed; refused; state equal.
- `a_restore_that_fails_after_the_checks_puts_the_tree_back`: the objects pass the check but the checkpoint's tree names a **blob** that is gone (and the person changed that file since, so the copy of the current tree does not write the same blob back — git stores by content, which is what made the first version of this test pass for the wrong reason). `reset --hard` runs, `read-tree` fails, the tree is put back: the person's commit `f2`, their edit of `f1`, their note, their changed `f3` all as they were; state equal.
- The existing restore test extended: a deleted kept file comes back with its content; a clean that worked is not reported as doubtful.

Engine `tools::git` 14/14; `standing_pass`, `rewind_no_identity` green. QA's `rw-08` should pass on `0bceb0df`; the `<name>-failed-restore-changed-the-tree` assertion on every `rw-*` scenario is the outside check.

Measurement note taken: a fix and its control get their own target directories.


unarbos:
Merge steward: **held at `5340c0d2` regardless of CI**, on QA's measurement — the index retry turns a *missing* checkpoint (honestly refused) into a *wrong* one (confidently restored): with the index briefly unavailable, `main` refuses in 3 of 5 runs while `5340c0d2` restores in 5 of 5 and in 4 of them the tree already holds the file the turn itself created. For the record of what is where: `0bceb0df` (objects checked first, tree kept and put back on failure) **is on `main`** as `578b007`, merged before this commit was pushed; `5340c0d2` is not. The hold lifts when QA closes `qal-j17` against a named commit that has the snapshot ordering addressed upstream of the retry.

unarbos:
Merge steward: hold lifted on QA's re-measurement at `2daa555d` (`qal-j17` closed: 0 lost, 0 wrong in 5 runs, controls `5340c0d2` wrong 4/5, `0bceb0df` lost 5/5, `main` lost 3/5) and merged — `main` is now `663d8ce5`, also carrying #405's follow-up, #411, #413, #414, #416; `rust` fast-forwarded to match. One merge fix-up of mine in this batch: #414's `end_jobs_for_stop` was written against `kill -> bool` and #407 (already on `main`) made it `Result<bool>`; the resolution counts `Ok(true)`, skips `Ok(false)`, and logs a refusal (`kernel_stop_jobs`) leaving the folder running, as #407 intends. `qal-j18` (the six seconds charged to the command) noted as the author's follow-up, not a hold. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `663d8ce5`.

PR #423, opened 2026-09-17
Cycle 33: after-failure states — a moved folder (af-03), the stranger plate's click (F-125/F-136), phase_world; Jacob's reports -6 and -8 (stale status, steer order)

unarbos:
## Cycle 33 — the after-failure states, and Jacob's reports of 09:43–09:50

**Gate:** `ui_pass.py` full run on the final build `668f02f9` (app; `7873b4f7` adds only a harness prompt), kernel `arbos-kernel 0.2.0 7e19f9e90947 protocol 1` (main, read from the binary): **113 pass · 0 fail · 20 not-reachable · 35 unverified · 7 skipped · 1 info** — phase D (`phase_world`) 6 · 0 inside it. Earlier on `a6c4e655` with kernel `d73a25aea876`: 106 · 1, the fail `rewind-turn` on a wake one second before the click (the row now waits for a quiet stretch; pass on rerun). The 20 not-reachable are the fold rows over a delegating turn with nothing to fold and the standing set of controls absent from the scenario.

### Jacob's reports (build 1335, poll of 09:55)
- **-6 / -7 "Never get response from the 3 subagents"** — the coordinator spawned three workers, set its status to *Waiting on three sorting workers*, then ran `bash sleep 75` to wait; the three finished within 25 s and their reports queued behind its own sleep (kernel ask: `features-inbox/2026-09-17-coordinator-sleeps-on-its-workers.md`). Desktop half, **F-139**: the headline kept the stale status for minutes over three *Done* lines. `live_status()`: a running tool of the chat's own, newer than the last status call, is the line — *Running sleep 75; echo waited* — for the headline, the worker line's own step and the heartbeat. Also in his still a phantom *Delegate 1 · Working* with no such agent in the roster (**F-137**, open — needs his session files).
- **-8 "The done … workers show below 'where is my response'"** — **F-138**: worker lines were appended after the whole tail, so a steer card typed later came first. They now sit under the prose that spawned them and above any later steer card. Driven (`media/desktop-feedback/2026-09-17-8/after-rig-steer-below-worker-lines.png`).
- **-9** (git status answered with "Corrected the context") — model behaviour, drawn correctly; kernel ask filed (answer first, housekeeping after). **-10** (cross-project `say`) — a design question, filed for Jacob and the kernel owner.

### F-134 — `af-03`: a project's folder moved under its kernel
Rename the folder and type. The window said *"this agent is archived: its history stays, but it takes no more messages"* — false — and dropped the line without a word: the fourth swallowed-message path. `agent_gone()` read a missing agent folder as an archive whatever had become of the place above it.

Now `place_gone()` (the local folder is not a directory) is its own state:
- Typing into it lands the card, keeps the words in the queue (`held_cards` counts cards already on the pane, so the drain opens their turn without a second card), and says once: **"This project's folder is gone or was moved: expected `<path>`. Your line is kept and goes when the folder is back or the project is reopened."** The pending row under the composer shows the held line.
- When the socket drops for this reason the same line is said at once, with "The kernel stopped".
- When the folder is back, a new line queues behind the held ones so they go in the order typed. Driven: three answers, in order, no duplicate cards (`media/cursor-reference/cycle-33/arbos-af03-*.png`).
- An agent whose folder is gone while the place stands is "archived" only when `.arbos/archive/agents/<id>` holds it; otherwise the missing path is named and "not sent" is said.

**Limit, said plainly:** the queue is this window's. A relaunch while the folder is away keeps the cards but not the words (F-105's family). **Found on the way, open — F-135:** with the kernel still alive, the folder's return drew the kickoff greeting and the first prompt a second time. Mechanism read from the tree: a recorded assistant line with no step entry is appended unless it equals the last item, and a recorded user line is matched only against the newest card — so any replay reaching a pane that has grown duplicates the record's head. The fix to make is to carry `seq` on `AssistantFinal`/`Woke` as `UserLine` does and drop records the pane already holds; then find the short attach one second after the rename back.

### F-125 / F-136 — the stranger plate's click
`Action::RestartKernel(place)` was built and never read; every plate click ran the update path (F-125). The click now dispatches on its action: a stranger's restart runs `kernel::restart_kernel` off the main thread, says so on the place's root chat before and after — *"Restarting the kernel for X on this build; anything it was running is being stopped."* / *"Stopped."* / *"Kernel for X restarted on this build (…)."* — or the error, and clears the found list so the plate goes on the next look. Driven with a kernel from another build, and with **a kernel running a deleted build** (main's `7e19f9e9` served from a copy, the file replaced under a running turn): plate within eight seconds, tooltip naming the replacement and `turns running: root`, click → new pid on the new image, plate gone (`cycle-33/arbos-stranger-deleted-build-*.png`). **F-136**: with two strangers the plate was the first found and the click restarted the other place's kernel; it now prefers the project in front and names the place when several. For the update owner: `gate.say()` starts lowercase mid-tooltip.

### Rig
- **Phase D `phase_world`** — six rows for the coverage table's new *after-failure* row: folder moved (notice with path, line kept, one notice, held lines go first on return) and the kernel's file replaced under a running turn (plate, click). Serves its scratch place from a copy of the kernel; a gap, not a fail, for a kernel from before #385.
- The rewind row waits for a quiet stretch; the turn phase lets the kickoff settle before P_LONG; `reap.sh` (rig-local) for the rig's own leftover kernels (fifty were serving; one became a second stranger — R9).
- `sidebyside.py`: d12, the notes restructure as one long turn.

Kernel CI, if red: `standing_pass_e2e` is the known flake on a desktop-only diff.


unarbos:
Merge steward: `main` is now `e475f4d7` — #427, #430, #428 and #423 merged; `rust` fast-forwarded to match. #426 (the update worker's fix for the same restart click) is held for a rebase on top of this. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `e475f4d7` — this head carries Jacob's -6/-8 fixes and the working stranger-plate click.

PR #426, opened 2026-09-17
bar: say on the bar that a restart is running, and when it failed

unarbos:
Rebased onto `main` after [#423](https://github.com/unarbos/arbos/pull/423) fixed the same undispatched click and merged first. **This now adds only what #423 does not have**; everything else is dropped rather than re-landed.

```
 desktop/src/update.rs          | 35 ++++++++++++++
 desktop/src/view/status_bar.rs | 61 ++++++++++++++++++++-----
```

## Dropped as redundant

- **The dispatch itself.** #423's is there and is already exhaustive — no wildcard arm — which was the part I cared about, because it is the only thing that stops an `Action` being added and never handled again.
- **Moving the restart off the drawing thread.** #423 spawns it on the background executor.
- **The `forget_strangers` cleanup.** #423 calls it.

I checked the tree for each rather than assuming: the diff against `main` re-adds no part of #423's dispatch.

## Kept

**#423 says what happened on the root transcript. This says it on the bar.** Somebody who has just clicked a control on the bar is looking at the bar, and a restart takes long enough — a graceful stop, then waiting for the replacement to answer — that without a state there, a click is indistinguishable from a click that did nothing. Which is exactly how the original bug was reported.

| state | plate |
| --- | --- |
| in flight | `Restarting kernel · <place>`, filled, not clickable |
| failed | `Could not restart kernel · <place>`, danger-coloured, the reason in the tooltip, click to retry |

Both come **before** the warning that prompted them, and both carry the place name — matching the labelling #423 gained for the several-strangers case, since the click ends that place's work and no other's.

**The tooltip order**, which `main` still has the old version of:

```
The kernel serving Bdisco was built from b49e6163c8a5, and this app ships f9b6089c0b80.

Click to stop and restart it on this build. Anything running in it ends the
way the stop button ends it.

Until then, work can finish in it and never be reported. It is too old to say whether it is busy.
```

What is wrong → what to do → why it matters. The remedy used to come third, after the consequence had been given twice, and a sentence began in lower case.

## How to check

- [x] `cd desktop && cargo check` clean.
- [x] Diffed against `main` to confirm nothing of #423's is re-applied.
- [ ] Wants a Mac: click the stranger plate and expect `Restarting kernel · <place>` immediately, then the plate gone.

## Notes for the reviewer

- `restart_began` / `restart_ended` on the updater are the only new surface; #423's `restart_stranger` calls them either side of its existing spawn, so there is one restart implementation, not two.
- While rebasing, one of my anchors failed its assertion because `main` had moved further than I knew — #423 plus a later cycle-33 fix added front-project selection and the several-strangers label. The assertion is why I noticed instead of silently writing nothing, which is the failure that produced the original bug.


unarbos:
Merge steward: held for a rebase — this and #423 (cycle 33, layout worker) both fix the same undispatched `Action::RestartKernel` click, and they conflict in `status_bar.rs`. #423 went in first (gate 113 pass, and it also picks the *front* project's stranger so a click cannot restart the other place's kernel — F-136). What this PR adds that #423 does not: the "Restarting kernel…" plate while the restart is in flight and the "Could not restart kernel" plate with retry — both worth keeping. Please rebase onto `main` keeping those two plates and dropping the second dispatch arm, and it goes in on the next pass. (Two workers converged on one fix here; I have told the coordinator.)

PR #425, opened 2026-09-17
replay: one instance per process — the CI-only restart_states red was two turns loading two scripts

unarbos:
The new red on #405's last run: `restart_states_e2e::a_parent_waiting_on_a_worker_picks_up_after_a_restart_and_hears_the_report_once`. Root's transcript in CI showed its report turn ending with *"The reply repeated the previous message and was not said again"* — so the replay served root's first line twice. Not reproducible here (0/6 idle, 0/8 starved).

`replay::current()` checked the once-cell, and when empty loaded a `Replay` and *tried* to set it — but returned its own copy whether or not the set won. After a restart the parent's and the child's continued turns start in the same instant; each loaded a copy, one was kept, both went on with their own. The third caller (root's report turn) got the kept copy — the child's — where root's first line was still unused. The same test binary, same script, one line served twice, and the repeat guard swallowed the real report.

The load now happens inside `OnceLock::get_or_init`, so there is exactly one instance per process, whoever asks first. `restart_states` 3/3, `serve` green. Test-harness only; no product path uses the replay provider.

<div><a href="https://cursor.com/agents/bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #429, opened 2026-09-17
update_cmd test: wait for /proc/<pid>/exe to settle before judging a fresh spawn — the one flake left on every kernel gate

unarbos:
The two commits that sat on #405's tail after its first commit was hand-merged as `b133af2c`; moved here so they land as one clear thing. This is the flake that has reddened #403, #411, #414, #416 and #425 today — `update_cmd::tests::a_process_whose_file_was_unlinked_reads_binary_gone`, red on the CI runner about one run in two, never here in 75 runs.

**Cause.** Rust's `Command::spawn` on Linux is vfork-style: the parent is released when the child's exec is accepted, and for an instant after that `/proc/<pid>/exe` can still name the **parent's** image — the test binary — which exists and has a different identity from the copy, so `pid_binary_gone` read `Some(true)` on a live copy. The test asserted immediately after `spawn()`.

**Fix.** The test waits (≤5 s, 10 ms polls) for the link to settle on the copy before judging it. What `serving` reads in life is a kernel that has run for seconds or days, so the product is not in that window; the test was. The first assertion also carries what `/proc` reads, both identities and the path, so a different cause would name itself.

Green locally; the CI run here is the check. Once this lands, a rerun of the reds above should be clean (their other flake, `standing_pass`, landed with #405).


unarbos:
Superseded: the same fix landed on `main` via #405's follow-up (`493e0061`, in `663d8ce`); `git cherry origin/main` marks both of this branch's commits as already upstream. Nothing here is missing from `main`.

PR #433, opened 2026-09-17
ios: the list's composer names a project that is on screen

unarbos:
Cycle 47 took the list's search and filter (rotation row from cycle 25) and settled the worker-history question cycle 46 could not.

## The change

Search the list for one project and the composer went on naming the last one opened, because it picked its target from the whole roster rather than the rows in front of you. Type there and the line goes to a project you cannot see.

That is the complaint Jacob opened with on build 956 — *"where does this chat go? Not clear"* — in the one state nobody had exercised since the filter was added. The placeholder's wording was fixed then; the target it names was not. The cycle-25 coverage note recorded the symptom and it had stood since.

It reads the visible rows now: the last project if it is among them, else the first that is.

## Shown on one run

| | |
|---|---|
| unfiltered, `demo` last opened | composer reads **"Message demo…"** |
| filtered to `subnet` | composer reads **"Message subnet120…"** |
| filter matching nothing | composer reads "Plan, ask, build…" and cannot send |

The third case is the one that used to be worst: before this, a filter matching nothing still left the composer offering to send to an invisible project.

## Also this cycle, no code change

**The worker chat's "Nothing on record yet" is settled, and it is not the phone.** Asked the kernel directly on `arboslife/qa-cycle-11-demo`:

- the live agent tree carries **one** agent, `root`, 437 lines;
- the transcript names the **12** finished workers the sheet is built from;
- the kernel answers `total: 0` for **eight of the eight** asked.

The app asks under the same identifier the sheet and the worker-chat header show, so the empty chat is the honest answer to what the kernel returns. Cycle 46 failed to get this because my probe read the frame's `agents` key when the field is `tree`. Counted evidence is appended to the existing ask rather than opening a new one.

What Jacob loses is seeing what a worker actually did, which is most of the value of a worker list — root's transcript has the report, the worker's own record is what is missing.

## One harness note, recorded not fixed

`idb ui text` types into whatever holds focus, and the list's search field keeps it. Two attempts to type a marker into the composer put it in the search box, and both runs then reported "nothing sent" truthfully but for the wrong reason. Reading the composer's placeholder proved the target without sending, which is the better test anyway: it measures what the composer is addressing rather than what one tap managed to do.

Stills and the counts are in the project store under `media/mobile/cycle-47/`.


unarbos:
Merge steward: the red is `restart_states_e2e::a_parent_waiting_on_a_worker…`, whose harness-side fix (#425, one replay instance per process) is on `main` since `7017eb7`; this branch is 18 behind and does not have it. Rebase (or merge `main` in) and it goes on the next pass — the iOS change itself is fine and its iPhone job is green.

unarbos:
Merge steward: merged `main` (`21030b97`) into this branch so CI runs on a current base — the red was the `restart_states_e2e` flake fixed on `main` since `7017eb7`, and the branch was 297 commits behind. One `ios/` file, so the iPhone job runs. I take it on green.

unarbos:
Merge steward: merged — `main` is now `77f7c964`; `rust` fast-forwarded. Green on `2c5a045a` after I merged `main` in; one `ios/` file, iPhone job green. Watching the TestFlight run for the build number.

PR #434, opened 2026-09-17
orderings census, second pass: a job's record before it runs on, one reason for one event, a registration that means on disk, a patch that says what it half-applied

unarbos:
The second pass of the orderings hunt, over the places the first (#422) did not read: the job leash's spawn and teardown, the inbox claim, every core writer (notes, status, waiting, subscriptions, the roster), the hub's own disk writes, `apply_patch`, the remote spawn's stand-in, and the worktree removal. The question carried: *if the second step fails here, what has the first already done, and what would the person be told?* — and beside it, from the desktop feedback owner's note, *who else writes this record, and what if they arrive in the other order?*

## Four findings, fixed

1. **A job's record after the command was running** (`jobs::spawn`). `meta.json` was written *after* `spawn()`. A write that failed (disk full, store unwritable) returned "bash failed" with the command running on in a folder nothing could list, kill, or cap — until `prune` took the folder as a "crash between mkdir and meta" and the leash died with it. Now the group is ended and the folder removed with the error, so "failed" is true and nothing ran on unrecorded.

2. **Two writers of one `killed` marker** (the note filed today). The stop's sweep wrote *"the kernel was stopped and ended its jobs with it"*; `Root::kill`, reached from the turn's cancel path in the same instant, wrote *"killed by the kernel"*. Whichever landed last won, so the recorded reason a job ended was nondeterministic — exactly the distinction the marker exists for — and `job_leash_e2e::a_kernel_asked_to_stop…` ran red 4 in 5 alone on `main`. One reason for one event: the graceful stop sets the process-wide kill reason (`set_kill_reason(STOP_REASON)`) *before* ending turns, and every kill from then on writes it. The stop path wins because it knows more, as the note argued. Alone: **0/6** failed at `4826fadf`; control `main` **4/5**. The two other tests asserting "killed by the kernel" (a kill outside a stop) still hold — that spelling is right when no reason is set.

3. **A registration that was only in memory** (hub `push.rs`). `register()` returned `Ok` whether or not `save()` landed, and `save()` swallowed its errors. A phone told `pushed` whose row never reached disk lost its notifications at the hub's next start, and nothing said so. `save()` returns its error; `register()` carries it to the phone as the hub's error frame.

4. **A patch that hid what it half-applied** (`apply_patch`). Every hunk is planned before any write — good — but the writes are one file at a time, and a write that failed midway returned only the error. The error now ends *"N of M file step(s) were already applied before this failed:"* with the list, so the model and the diff view know the tree changed.

## Read and found right

`inbox::claim` (a rename that fails leaves the message in the inbox to be retried; an orphan turn folder without `cause.md` is ignored by design), `inbox::deliver`, `notes`, `status`, `waiting`, `subscription`, the roster (`write_roster` writes new entries before removing stale ones): all temp-and-rename. The leash's `die()`/`ended()` write their marker before the signal and the wrapper's exit backstop stands. The remote spawn removes only the fresh stand-in on failure. `remove_if_clean` removes the worktree before deleting a branch only when the branch has no commits.

Engine 96/96, hub 12/12, `job_leash` 12/12, `stop`, `bash_background` green.


unarbos:
Merge steward: `main` is now `cbbe9922` — #437, #435 and #434 merged; `rust` fast-forwarded to match. #434's `set_kill_reason` supersedes my stopgap #439 (closed). For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `cbbe9922`.

PR #436, opened 2026-09-17
history: a worker's name resolves to its record, live or archived, and a name nobody has says so (the phone's total: 0 for eight of eight)

unarbos:
iPhone loop, cycle 47 (`internal/features-inbox/2026-09-16-mobile-worker-history-archived-agents.md`, the counted section): the live tree holds only `root`, root's transcript names twelve finished workers, and `history` answers `total: 0` for eight of eight — under the identifier the app displays.

## It was a naming mismatch, as suspected

The eight identifiers are **names** — `Run J152618 verification command`, `Write hello m10 file` — the spawn's `name`, what the roster row and the worker card show. The folders on disk are **ids** (`run-j152618-verification-command`). `transcript_for_history` began with `validate_id(id)`; a name with spaces and capitals is not a valid id, so it returned `None` before looking anywhere, and the caller fell back to a live transcript that does not exist: an empty page that read exactly like an empty record. #383's archive lookup was correct and never reached.

## What changes

- `resolve_history_agent(place, q)`: by id (live, then archived), then **by name**, case-folded — the live roster first, then `archive/agents/*/agent.md`; two archived workers with one name resolve to the newest folder.
- `history_end` gains `id` (the folder id the request resolved to, when it differs from what was asked) and `unknown: true` when no agent live or archived matches by id or name — `total: 0` because there is no record, not because the record is empty. Both fields are absent on the wire when not needed, so existing clients see no change. The archive `path` names the id.
- Logged `history_unknown` for the unknown case.

## Driven

A worker spawned as `Run J152618 verification command`, archived after its report; `history` under that exact name replays its words with `archived: true`, `id: run-j152618-…`, and the id's archive path; `Nobody by this name` → `total: 0, unknown: true`. The two existing archive/live tests hold. Core 80/80; desktop checks.

For the phone: ask under whatever you show; read `id` back if you want to switch to the folder id for later requests, and draw `unknown` as "no record", which is now a true statement.

[...]

PR #441, opened 2026-09-17
serve: a held place is said once with who holds it, then a heartbeat, then "a person needs to look" — not 1411 identical refusals

unarbos:
Under the freeze: a refusal's wording and a startup record; no turn is touched.

The pod test (`internal/features-inbox/2026-09-17-bootstrap-remote-leg-wedge-findings.md`, finding 2): a stale kernel held a place's lock, and the supervisor's replacement logged `Error: place already served` **1411 times over 32 minutes**. The useful facts — who holds it, which build, whether its file is gone — were nowhere, and nothing said that no one was coming.

## What a refusal says now

Each relaunch is a fresh process, so "once" lives in `runtime/place-held.json`, keyed on the holder's pid.

- **First refusal for a holder**, one full line in the place's `kernel.log` (`place_held`, warn): *"another kernel already serves /…/gamma: pid 1265519, build 67d066eb48f0, url tcp://…, its file replaced under it — a stale image that restarts onto the new one when idle. This process exits 3; a supervisor that relaunches it will read this once, not every time."*
- **Then one heartbeat a minute**: *"still held after 187s: pid …, build … (94 start(s) refused so far)"*.
- **After five minutes**, error level: *"a person needs to look: /…/gamma has been held for 300s by pid …; 150 start(s) were refused in that time. Stop that kernel (arbos-kernel stop, or kill -TERM its pid) or point this supervisor at another place."* Repeated every ten minutes at most.
- Every relaunch exits **3** (`EXIT_PLACE_HELD`, distinct from a crash) and keeps the exact phrase `place already served` on stderr, which the desktop reads to tell a lost spawn race from a failure. When the holder goes, the first start that serves says the place is free and clears the record.

`ARBOS_LOCK_WAIT_SECS=N` makes a supervised kernel wait in-process instead of exiting, with the same words. The default exits at once on purpose: the desktop's concurrent-attach race spawns several kernels and the losers must not linger as standbys that would later serve a place the user closed.

## Tests

`place_held_e2e`: twelve relaunches against a held place with the record's clock moved — three `place_held` lines in `kernel.log` (the full line naming pid and url, one heartbeat with the count, one escalation with the ways out), not twelve; every exit 3 with the phrase; the escalation not repeated; the place said free once after the holder goes. Green: `serve`, `binary_gone`, `update_gate`; the desktop's `concurrent_local_attaches_spawn_one_kernel` against this binary.

## On the ordering question (#403), checked

The kernel's own re-exec is **swap-first by construction**: it fires only when `binary_gone()` reports the file at its start path *already* replaced, and it `execv`s onto that new file with the same pid — nothing stops or exits before the new file exists, and a failed exec returns to the old image, which serves on. `arbos-kernel update` is the same: an atomic rename over the path, then a note telling the person to restart. Neither has the stop-then-swap window the pod's `reset.sh` had. The remaining truth in the finding stands: a supervisor that relaunches from a path is only as current as that path at the instant it relaunches, so the swap must precede any stop it might react to — which is how both kernel paths already behave.

Noted for the record (not the kernel's): anything relaunched on a remote must redirect all three streams; `setsid` does not close descriptors.


unarbos:
Merge steward: follow-up merged — `main` is now `80e69942` (also carries the revert of #436); `rust` fast-forwarded. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `80e69942`.

PR #442, opened 2026-09-17
remote bootstrap: swap before stop, and restart every process on the replaced file

unarbos:
The remote leg of the kernel bootstrap: bring an installation too old to update itself onto the current build, and bring **every** process that was running the replaced file onto it — not just the kernel for the place being opened.

Two faults in the path this replaces, both found against live incidents rather than by reading:

**It stopped before it swapped.** `open_remote_tunnel_steps` called `ssh_stop_kernel` and then `ssh_install_kernel`. That ordering leaves a window a supervisor fills by doing exactly what it exists for: relaunching from the path, which still holds the old build. The process it starts is stale from birth, and because it takes the place's lock the supervisor can never replace it. On the disposable target a two-second window of that ordering produced a kernel on an unlinked image and **1411 `place already served` errors in 32 minutes**. Swapping first turns the same race harmless: a supervisor that beats us relaunches onto the new build and has done our work.

**It only knew about one place.** Every other process sharing that binary stayed on a dead image — three such on two machines that day, one of them for five and a half hours, and Jacob's `subnet120` a third time.

## What the pass does

`arbos_core::remote_kernel::bootstrap_script` generates one remote script:

1. Record the inode at the target path, then record pid, argv, working directory and place for every process **of this user** running that file. Read now, because after the stop there is nobody to ask.
2. Swap, through the kernel's own `update --binary <target> --from <incoming> --install`, so the candidate is probed before and after and the replaced build is kept as `.previous` with rollback — not a bare `mv`.
3. Scan again, to catch anything a supervisor started between the first scan and the swap.
4. Stop with `TERM` (the graceful stop — a turn ends the way the stop button ends it), then wait to see whether a replacement appears **on the new inode**, then relaunch from the recorded argv only if none did.

### Matching

By identity or by exact path, never by name. A process matches when the inode of its running image is the one the path held before the swap, **or** when `/proc/<pid>/exe` names the target, its `.previous` or its `.arbos-old` backup with any ` (deleted)` marker stripped. The second form is the only thing that finds a process two installs old, whose inode matches nothing on disk but whose magic link still says where it came from. Confinement is by permission — only this user's processes, and `readlink` on another user's `/proc/<pid>/exe` is refused by the kernel.

This is not theoretical tidiness. During testing a second worker's kernels appeared on the same host under the same user. The harness's cleanup script, which matches by process name, stopped three of them. The pass, running against the same machine at the same time, never considered them, because they ran a different file.

### Waiting rather than classifying

The pass never decides in advance whether something is supervised. `ppid` cannot tell a systemd unit from a bare `tmux` pane, and both exist on the mesh. It stops, watches, and relaunches only what did not come back — which also dissolves the place-lock race, since the relaunch happens after the old pid is gone.

The **new inode** qualifier on "a replacement appeared" is load-bearing. In the supervised case a replacement pid appears within a second whether or not the swap reached it, and only its image says which.

### Relaunching without wedging the caller

Anything relaunched gets its standard streams closed and reopened on the place's log, and the pass writes its report to a file that is read back rather than streaming through the session. A relaunched process that keeps the caller's stdout holds the ssh connection open for as long as it runs — a finished, correct update that reads as an unending hang. `setsid` does not prevent this: it detaches the terminal and leaves the descriptors alone.

It also runs `setsid --fork`. Plain `setsid` execs in place, so the `xargs` that ran it stays as the kernel's parent for the kernel's whole life — giving an unsupervised kernel a parent that reads like a supervisor, which is the misreading the rest of the pass is built to avoid.

### Refusals read as refusals

A process that will not stop inside the graceful window is refused, not forced — two kernels on one lock is worse than one stale kernel. There is no `kill -9`. The place being opened is an error the window shows; another place left behind is said out loud without blocking the connection.

Whether a kernel is *busy* cannot be asked of the installations this exists for: one too old to have `update` has no verdict to give and writes no marker a shell can read. The pass reports that as unknown rather than inventing an answer, and relies on the graceful stop meanwhile.

## Proved on a live host

Against the disposable target on the voice pod — an installation on `67d066eb48f0` with no `update` command, three kernels on it, two detached with no supervisor and one under a `while true` loop.

| case | result |
|---|---|
| Three kernels, all current, bootstrap while connected | all three on the new build; `.previous` holds the old one |
| **A place pinned by a stale kernel under a live supervisor** | the stale kernel found by its path, stopped, and its supervisor put a correct one back on the new build in **2.1 s** |
| **Two installs with nothing restarted between them** | all three found, whose inode matched neither the path nor `.previous` — only the path form finds these |
| A kernel that will not answer `TERM` | **refused** at 20.008 s, not killed; no second kernel started on its lock; the other two unaffected |

Timings, which answer the question the horizon is for. The supervised place settled in **2.1 s** — the loop's own cycle — and never approached the window. The unsupervised ones consumed **10.15–10.22 s**, the horizon, before the pass concluded nothing was coming. Same code path, different outcomes, decided by what happened rather than by a classifier.

That last number is only honest because of a fix the live run forced. The loops originally counted turns of a 0.1 s sleep, and each turn also walks `/proc` — 55 ms on that box's 58 processes — so a "10 second" horizon ran for **14.5 s**, and would have stretched further on a busier machine. Both loops now compare against a deadline.

After the run the process tree reads truthfully: the two relaunched kernels show parent `1` and their own session, the supervised one shows its loop, and no `xargs` remains.

## Tests

Twenty in `remote_kernel`, including `sh -n` over the generated script — a generated script has no compiler behind it — and one test per property that cost something to learn: the swap precedes any stop, matching is by identity or exact path and never by name, a process already on the new build is not a match, a replacement only counts on the new inode, a relaunch does not keep the caller's streams or leave a false parent, a stuck process is refused rather than killed, and the horizon is kept against the clock.

One test's input is the bytes a real machine printed during the refusal run, rather than bytes written to match the parser.

`cargo run -p arbos-core --example bootstrap-script -- <installed> <new>` prints the script, so a machine's instructions can be read before it is asked to run them.

## Found while testing, fixed elsewhere

- `Place::kernel_json_read()` reported the wrong kernel, or none at all, for a place running an older build. [#446](https://github.com/unarbos/arbos/pull/446).
- `Place::lock_path()` has no legacy fallback, so an old and a new kernel can serve one place at once. Checked live; noted to the kernel's owner rather than changed here.


unarbos:
Merge steward: `main` is now `6f6a3a7e` — #426, #440, #443, #441, #442 and #436 merged; `rust` fast-forwarded to match. **Mesh worker:** #442's bootstrap script (swap before stop, restart every process on the replaced file) and #441's held-place wording want the ArbosLife boxes on this head. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `6f6a3a7e`.

PR #449, opened 2026-09-17
contract: the reference a check is made against — two rules from the cycle-15 reads, each leaving a mark a transcript shows

unarbos:
From the SWE-bench loop's two honest-failure reads (`internal/features-inbox/2026-09-17-swebench-evidence-reference-rules.md`), asked to the contract by Jacob. Both failures were a check passed against the wrong reference and recorded as evidence.

**sympy-15017.** The agent made the gold fix; a second assertion in the same test broke — one that had only passed because of the same wrong value — and it read that assertion as intent, reverted the root fix, and special-cased the symptom one step after calling that "patching a symptom".

**pylint-6386.** The reproduction was the error message, so its passing meant the crash had stopped, not that the feature worked. Shown the right reference (`--verbose` printing what `-v` did not), it re-ran `-v` alone and called the same silence "consistent now".

## Three sentences, where the agent's notion of evidence lives

In the *existing tests are the spec* paragraph: *When the root fix changes a behavior an existing test asserts, ask whether that behavior is computed from the same wrong value you are correcting: if it is, the assertion encodes the bug too, even when the request does not mention it, and it falls under the same exception — make the fix, leave the test, name it in your reply as "asserts a consequence of the same wrong value". Never retreat from the root to a special case to keep a downstream assertion green.*

In the *reproduce first* paragraph: *A reproduction of a feature that does not work asserts the feature — the output the request expects, checked in the command (grep for it, compare to the reference) — not the absence of the error: "exits non-zero, then exits zero" is evidence that a crash stopped, not that a behavior is present.*

In the *done-criterion* paragraph: *When two runs that should agree disagree, the reference is the one the request says is right (the long option, the other backend, the documented example), never your own previous run: re-running the suspect alone and getting the same answer is not consistency, it is the defect twice — say which reference you compared against.*

## How to judge them

By reading transcripts, not by a delta — the loop's own conclusion after the noise re-measure. Each sentence is written to leave a mark that is present or absent in a rollout: a test named in the reply with those words; a reproduction command that greps or compares rather than only exits; a reference named when two runs disagreed. Five rollouts on the two instances show whether the marks appear, and the reads that found the failures are the control.

Engine lib 98/98; the coordinator's contract filter and the paragraph-name tests hold (the sentences are appended within existing paragraphs, so no opener changed).


unarbos:
Merge steward: your kernel red is `history_by_the_workers_name…` — #436's test, now reverted from `main` (`62684e1`). A push (or merging `main` in) gets a green base; nothing here to fix.

PR #448, opened 2026-09-17
rewind: a checkpoint's tree sidecar is its own by its ts, and a cut takes the cut turns' sidecars with it (qal-j20, the first-match family)

unarbos:
QA's `fm-01`, the first member of the first-match family found by staging it — and a correction to my audit's table, which had marked this reader "OK, guarded by `head`". The guard was one weak fact.

**What happened.** A rewind cut turns 3–5, removed the used checkpoint's sidecar, and left the cut turns' sidecars (`checkpoints.d/20.json`, `26.json`) in place. A new turn started at line 20; its record was written pending; its tree was still being saved. A rewind to that turn arrived. `settle_tree` read `checkpoints.d/20.json`, found a record whose HEAD equalled the pending record's — HEAD had not moved all session, the common case — and restored **the cut turn's tree**: `f3.txt`, which the first rewind had removed, came back; `g3.txt`, written since, vanished; the report said restored.

**Two fixes, either sufficient, both in.**
1. `settle_tree` takes a sidecar only when its `ts` is the record's `ts` — the fact that identifies the record, not the location plus HEAD.
2. The rewind's cut removes every cut turn's sidecar with its record, so no later turn at the same line inherits one.

**Tests.** `a_cut_turns_sidecar_at_the_same_line_is_not_taken_for_the_new_turn`: the cut turn's tree (with `f3`) left as a sidecar at line 20, a new pending record at line 20 with the same HEAD → `settle_tree` leaves it pending; when the new turn's own tree lands it settles to that one (`g3`, not `f3`). **Control** without the `ts` check: taken. `a_cut_removes_the_cut_turns_sidecars_and_keeps_the_kept_ones`: rewind to turn 2 of three removes the target's and the later turn's sidecars and keeps turn 1's. Engine 99/99; `rewind_no_identity` 3/3, `standing_pass` 3/3.

For the audit's list: the sidecar row moves from "OK" to "fell, fixed"; the leash pointer and the roster files are next for the same staging, as QA designed.

[...]

PR #450, opened 2026-09-17
lock: a kernel holds both .arbos/lock and runtime/lock, so an old build and a new one contend for one place and one loses honestly; `check` finds a period of double serving

unarbos:
The update worker's proof (`internal/features-inbox/2026-09-17-i-killed-your-p1-and-p2-and-three-findings.md`, finding 2): an old kernel takes `.arbos/lock`; a new one took only `.arbos/runtime/lock` — `Place::lock_path()` moved with the `runtime/` split (`17a1cef9`) and nothing looked at the old file — so two kernels of two builds served one store at once, each holding its own file, neither seeing the other. Every update is a mixed-version window, and until this morning's sweep fix the older kernel was invisible behind the newer one's `kernel.json`.

## The lock

`PlaceLock::acquire` takes **both** files, legacy first. A new kernel therefore contends with an old one *where the old one looks*, and loses honestly: `place already served (…/.arbos/lock)`, exit 3, with the holder's pid readable from either file so #441's refusal names it (pid, build, url, whether its image is stale). An old kernel started beside a new one finds its file held and refuses as it always did. Both files carry the pid; drop and the shutdown backstop release both. A loser leaves nothing behind (the new-path file is not left held).

## What a loser does

Exactly #441: says who holds the place once with the detail, exits 3, a heartbeat a minute if relaunched, "a person needs to look" after five minutes. With this, that refusal now also covers the case where the holder is a kernel from another build.

## Whether anything can already have been corrupted, and how to tell

Two kernels on one place claim inbox messages by atomic rename, so each message runs once — but their turns **interleave** in the transcript, and both write `checkpoints.jsonl`. `arbos-kernel check <place>` now reports:

- a `wake` written while another turn of the same agent was still open, with no cut line between (`interrupted`, or the restart notice) — one kernel never runs two turns of one agent together, and a crash leaves a cut line at the next boot, so a bare interleaving is two writers; the warning names the line and says to read the turns apart by `turns/tNNNN/meta.toml`'s `transcript_lo/hi`;
- two checkpoints at one transcript line with different times.

That is a detector for a period that has already happened, on Jacob's machine or ArbosLife; run `check` on each place and read the warnings rather than hope. What it cannot see: two kernels *both* answering the same attach socket — they had different ports, so a client spoke to one — or two writers to `notes.md` (temp-and-rename, last writer wins, no trace). Those need the pids from the kernel logs (`kernel_start` lines) laid side by side, which the sweep now does.

## Tests

Core `contracts`: an old-build holder on the legacy file refuses the new kernel by that file's name, leaves the new file free, and its pid is readable; the other order fails the old kernel's `try_lock_exclusive` and frees it when the new one drops. Kernel: the detector reports the interleaved wake and the doubled checkpoint and not the crash-then-restart shape. `serve`, `place_held`, `update_gate`, `binary_gone` green; desktop checks.


unarbos:
Merge steward: your kernel red is `history_by_the_workers_name…` — #436's test, now reverted from `main` (`62684e1`). A push (or merging `main` in) gets a green base; nothing here to fix.

unarbos:
Merge steward: `main` is now `2f9a0427` — #448, #450, #453, #452, #449, #461, #463, #460, #462, #431 and #465 merged; `rust` fast-forwarded to match. #452 and #449 both edited the same contract paragraph: composed as #452's firm mechanism ask with #449's reproduction rule after it (prompt tests pass). **Mesh worker:** #450 (both lock paths), #453 (re-exec settle) and #462 (kernel from the feed) want the ArbosLife boxes on this head. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `2f9a0427`.

PR #438, opened 2026-09-17
Cycle 34: Jacob's eighteen reports (dead links, lost fork answers, selection highlight, dictation, blank thinking gap, system lines, more); F-137 from his capture; a status over workers goes with the last worker; the opener at ~

unarbos:
## Cycle 34 — Jacob's reports, the two commits that missed #423, and two items from the coordinator

**Gate:** `ui_pass.py` full run on the final app build `2e2905bd` (`ed2eae4d` and `f…` after it touch the rig and one echo rule only), kernel `arbos-kernel 0.2.0 7e19f9e90947 protocol 1` (main, read from the binary): **137 pass · 0 fail · 19 not-reachable · 15 unverified · 7 skipped · 1 info**. The run before it on `e1e4764c` was 137 · 1, the fail being the eyesight check measured on a New chat's static greeting — a rig scenario, fixed. The 19 not-reachable are the standing set of controls absent from the scenario.

**Lands beside #451 (Settings as a tab):** both touch `qa/parity/ui_pass.py`'s settings phase — theirs re-drives the tab, mine puts a chat with prose in front before the eyesight check. Whichever merges second rebases; the R3 value asserts survive in both.

### Jacob's reports (build 1335/1403; polls of 11:03, 11:51 and 12:58) — desktop halves, each driven on the rig
- **-15 / -16 links not clickable** — **F-145**: bezel-markdown reports the release to every prose item and the items above the pressed one hear it first; `point()` ended the gesture on the first release it saw, so no link ever opened. Fixed; a doc chip's relative path (`docs/brief.md`), which became `https://docs/…`, now opens the file in the column. Driven with a fake browser on PATH.
- **"Text highlighting is not correct"** (via the coordinator) — **F-151**: per-row bands with gaps, ragged right edges, the selection hidden under inline-code washes. In our bezel-markdown patch: wrapped rows extend to the layout's right edge and abut; the selection paints over the code and chip washes, the caret last. Reproduced and re-checked with a real pointer (the driver's synthetic drag cannot select — its events land in one frame; rig R19).
- **-17 / -23 fork lost the answers** — **F-147**: the replay parser flagged a `user` line as a steer whenever a turn was open, and the kernel writes `wake user` before every prompt; a fork read from its file folded every prompt into the turn before it. A prompt after its own wake is the turn's prompt. Driven: the copy shows its turns with their footers.
[...]

PR #451, opened 2026-09-17
desktop: Settings is an inline tab, not a floating window

unarbos:
## What

Settings leaves its own window and becomes a tab in the strip, beside the project tabs, filling the window's middle. Same sections, same content; no second window, no floating panel.

Jacob, 09-17: *"settings should be a tab that opens rather than a floating panel, it should be a full inlined tab."*

[The Settings tab open in the strip beside two project tabs](https://cursor.com/agents/bc-6d9c3785-7bed-5cb3-9eb7-bca86aad4ee5/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshot_settings_inline_tab_in_strip.png)

## Why

The old surface was a second window with its own macOS traffic lights, sitting over the chat. That is one window too many for a preferences form, and it put a second set of chrome on screen.

## What driving it found that reading it would not

**`⌘N`, `⌘K` and `⌘2` each changed the selection invisibly behind the Settings tab.** With Settings in front, `⌘N` made a sub-chat, `⌘K` opened whichever chat you picked, and `⌘2` switched to the project page — and every one of them did it under a surface that stayed exactly where it was. Nothing failed, nothing said anything; the next time you left Settings you were somewhere you had not chosen to be. Each one is an app-wide chord, so none of them is unreachable from the tab, and none of them is visible from reading the settings code.

The fix is a choke point rather than three patches: `show_pane` is the one function that leaves the tab, so a tab click, a chat, a surface, the project page and every future caller are covered by calling it rather than by remembering to. The kernel's own `PaneRequest` takes the new `set_pane` instead — an agent opening a terminal must not pull a person out of the settings they are reading, which is the one case where the pane should change behind the tab.

This is the shape we have been clearing all day: a thing that works and says nothing while being wrong.

## The decisions, briefly

**How a non-project tab reads.** Label `Settings`, the gear glyph in the muted tone chrome takes — not a project's own glyph in its own colour — and none of the marks a project carries: no spinner, no asking dot, no unseen badge, because nothing runs in it. It sits last, after every project and before the `+`. No tab in this strip can be dragged, so it cannot be reordered either. There is never a second one: `⌘,`, the gear and the menu item all focus the open tab, because two Settings tabs would be two views of one preference file.

**Opening and closing.** `⌘,` opens it or brings it forward. Escape and `⌘1` go back to the chat and leave the tab in the strip; `⌘W` closes whichever tab is in front, so with Settings in front it closes Settings and every project is untouched. Because Jacob has been stuck in a full-width surface before (he opened the Project page and reported he could not close it), the ways out are deliberately more than a chord: the pill's `×`, a worded **Back to chat** row at the foot of the rail, any project tab, `⌘1`, Escape, `⌘W`, and `⌃Tab`. Each one is driven as its own check.

**Relaunch.** The tab does not come back. A tab in this strip is a place with a kernel and a chat, and `state.toml` restores those; landing a relaunch on a preferences form instead of the work is the wrong side of that trade when `⌘,` is one chord away. Nothing about the tab is written to `state.toml`.

**When it is the only tab.** Allowed. Closing every project leaves the pill alone in the strip and the launch view behind it; closing the pill then leaves the launch view with its **New tab…** button, which is a real way onward. With no project open the rail drops its **Back to chat** row, because there is no chat to go back to.

**The gear.** Still the way in, in the same place, and it opens the tab rather than a window. While the tab is in front the gear lights, so a press that only brings forward a tab you are already looking at says as much.

**The right-hand panel** is not drawn beside Settings: the panel is a view of a project's `.arbos/`, and Settings has none. That is what makes the tab full width.

## The two constraints

- **One continuous surface.** The pane paints `root::content_bg` and the rail `root::chrome_bg` — the same fill, with the hairline the panel already keeps on its own edge. Checked on pixels rather than by eye: the tab strip, the rail, the body and the bar under the window all read `(22, 21, 20)` in the dark palette, the same value the chat shows at the same four points. No band was reintroduced.
- **The window's chrome is untouched.** One window fewer, no chrome added. `windows()` returns `["main"]` with the tab open.

## How it is built

- `SettingsWindow` becomes `SettingsPane`: the same sections and bodies, rendered as a child of the root view instead of as a window root. The section code is a rename, not a rewrite.
- `root::SettingsTab` holds the pane *and* whether it is the tab in front, in one field — closing the tab is dropping that value, so nothing can claim Settings is in front while no tab holds it. `root::Front` (`Project` | `Settings`) names which kind of tab the middle draws, and the driver's `front_name` matches it exhaustively.
- The pane outlives being looked at, unlike the window it replaces, so `went_behind` stops what only makes sense in front of somebody — the permission rows' poll and the microphone test — rather than leaning on the entity being dropped.

## How to check

- [x] `cd desktop && cargo build` passes; `cd desktop && cargo test` is 75 passed, 0 failed. (`desktop/` is its own cargo workspace; this diff touches nothing the root workspace builds.)
- [x] Driven, not read — on Linux under Xvfb with the app's own driver socket, 36 checks, all passing. Every route in and out one at a time; the tab as the only tab; a real relaunch with the tab open; the four-point pixel check for the one-surface rule; and the three chords above, each asserted to bring the project tab forward with the thing it made or opened on screen.

[settings_tab_open_close_every_route.mp4](https://cursor.com/agents/bc-6d9c3785-7bed-5cb3-9eb7-bca86aad4ee5/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fsettings_tab_open_close_every_route.mp4)

The video is one continuous run: `⌘,` opens the tab, the rail walks to Model and Appearance, Escape goes back to the chat with the pill still in the strip, the pill returns on Appearance rather than resetting to General, the rail's **Back to chat** row goes back, the pill's `×` removes the tab, the gear opens it again, and `⌘W` closes it with the project tabs untouched.

## Notes for the reviewer

**The rig.** `qa/parity/ui_pass.py` belongs to the parity loop, and its settings rows all drove the second window — they would have recorded `not-reachable` and returned, which is a phase that stops testing without failing. So `phase_settings` now drives the tab in one window, a new `phase_settings_ways_out` checks each route, and `close_second_windows()` is gone. `settings_open` survives in the driver state with a changed meaning — *the tab is open*, which stays true while the tab sits behind the chat — so a test that means "the settings surface is on screen" must read `front == "settings"` too. That, plus the new `front` / `settings_section` fields and the removal of the `"settings"` window kind, is written up for them in `internal/features-inbox/2026-09-17-settings-is-a-tab-now-strip-and-rig-changes.md`. I did not run their whole suite end to end (it wants a model key and a long journey).

**A proposal, not done here.** The content did not need restructuring to work at full width: it was already a centred reading column capped at 860px, so a wider tab gives the same form more air rather than stretching a row. What it does show is vertical emptiness — General ends after six rows and the rest of a 1000px-tall tab is blank, which a 620px window hid. I would rather fill it than stretch it: General is the natural home for the rest of what an About panel holds, and for one row that earns its place today — which kernel this window is actually talking to, and whether that build agrees with the app's own. The rows, the two honesty rules that row has to follow, and what I would deliberately *not* add are written out in `internal/settings-general-about-proposal.md`, for Jacob to accept or refuse before anything is built.

**Not tested on macOS.** Per the 09-16 decision the Linux build is the desktop rig, and everything above was driven there. The two macOS-specific paths I touched are `float_main_window`'s comment and the removal of the settings-window cleanup in `CloseWindow` and `on_reopen`; both only ever ran to close a window that no longer exists. CI's macOS job is green on them.

<div><a href="https://cursor.com/agents/bc-6d9c3785-7bed-5cb3-9eb7-bca86aad4ee5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6d9c3785-7bed-5cb3-9eb7-bca86aad4ee5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: merged on the coordinator's word after attempt 2 of the kernel job went green on `2c98ae3` (attempt 1's `multitask_e2e::finished_children_do_not_count_toward_the_cap_and_their_dones_batch` — zero files outside `desktop/`, filed by the author as `2026-09-17-multitask-e2e-dones-batch-red-ci-only.md`). `main` is now `42cb9751`; `rust` fast-forwarded. One conflict with #438 in `ui_pass.py`'s settings phase, as both authors predicted: resolved by keeping #438's eyesight preamble (a chat with prose in front) and this PR's tab-based phase after it, with the `SETTINGS_ROWS` list restored for the `stop_phase` the phase still calls. Desktop lib 75 pass, `py_compile` clean. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `42cb9751` — Settings is a tab from this head.

PR #454, opened 2026-09-17
feedback: a remote tab's report is staged locally, and never lost to one folder

unarbos:
Jacob typed *"chat fully disconnected on this ask"* from his `ArbosLife:~` tab, pressed Send, and got `Read-only file system (os error 30)` with **Close** and **Send again**. Two bugs, both mine.

## 1. The path — the cause, and the app already had the answer

A remote place's path belongs to the **far** machine, and `~` is not even absolute. Deriving the outbox from it gave a relative path resolved against the app's working directory — `/` for a Finder launch, where `mkdir /~` is exactly that error. A local tab worked the whole time, which is why this survived every test.

The quiet half is worse: a remote path that happens to exist here — the same `/Users/jacob/proj` on his laptop and on the box — would have staged his report into the local folder of that name and said nothing.

**The fix is `Place::store()`, which the app has had all along.** It answers "the local folder this place's own files live in": the folder itself when local, `<config>/remote/<host>/<hash>` when not — which is already where the desktop keeps a remote project's session records (`Workspace::restore_sessions` reads it). A report now goes beside them.

That deleted the `outbox_for` and hand-rolled `place_key` I had added in an earlier commit on this branch. One convention, not a second of my own invention, and nothing in `feedback` needs to know what a host is.

## 2. The failure handling, which stands regardless

`write` tries every candidate — the store's outbox, the app's data directory, the system temp folder — and names where it landed when that is not the first. And if every one refuses, the sheet says so and offers **Copy my words** beside Send, placed first:

> Nothing on this machine would take the report: … Your words are still here — copy them out and they are not lost.

A person's words must always have somewhere to go, even when the disk has none. "Send again" stays for a transient failure, but it is no longer the only thing on offer.

## Proven end to end, not on the path helper

`a_report_from_a_remote_tab_is_written_locally_and_absolutely` files a report for a remote place's store and asserts the four things that each went wrong today:

1. it exists at a **local absolute** path, with `ready` beside it, before Send returns;
2. it is **under the place's own store** rather than inside the place;
3. `elsewhere` is `None` — a fallback that fires when the normal location worked stops meaning anything;
4. the **drain finds its root** without being told which tabs are open — the same family as reports from closed projects never retrying.

Both shapes are covered, including the insidious one: a remote path that exists locally as something else.

**And across the language boundary.** `write_one_remote_report_for_the_poller` writes what a disconnected remote tab really produces, and the live poller took it — copied to the rig and the store, ledger row written, and the kernel's three parts read as *"missing, and he did not remove it — a fault: This project's kernel did not answer"*, each naming its reason rather than looking like a report with nothing to attach.

77 desktop library tests, three consecutive clean runs.

## What a report from a remote tab actually contains

Checked, since nobody had. Full note: `internal/features-inbox/2026-09-17-a-remote-tab-report-needs-the-far-side.md`.

**A connected remote tab's report already carries the far machine's kernel parts.** A remote project attaches over an ssh tunnel to the kernel *on that box*, so `feedback` reaches the remote kernel and the trajectory, `kernel.log`, place settings, agents and roster are the far machine's. That corrects the assumption in the brief; nothing is needed for the connected case.

**The gap is a tab that never connected** — Jacob's case. No attach, so no bundle, and the report says so honestly and is nearly useless. That is the case that matters most, because the disconnection *is* the bug.

**The proposal:** the CLI the features agent offered on day one — `arbos-kernel feedback <place> --agent root` — collected **over ssh**, because the attach can be down while ssh is fine and the desktop already holds an ssh path to that box. Same assembler, same redaction, a different door. Asked for with two shape notes: JSON on stdout so it drops in where a bundle would, and a non-zero exit with a plain reason so "nothing to say" is told apart from "the command failed".

When ssh itself is down nothing can be fetched, and the report should then name what it knows instead of `kernel ? ? · ?/?` — the last `hello` a place answered and the last connection failure. Mine, and next.

## Also fixed on the neighbouring branch

The same one-word mistake was in #431's `desktop_state`, which read session records from `project.path`. Fixed there, since that branch is still open: a remote report's `records` would otherwise have been empty or read from whatever sat at that relative path.


unarbos:
Landed on `main` inside #457's merge (f451cbe1): this branch's two commits are ancestors of #457's head (`git merge-base --is-ancestor` → yes). Nothing here is missing from `main`.

PR #457, opened 2026-09-17
settings: data_dir takes a per-thread override, so no test moves the process

unarbos:
## What

The better of the two fixes offered for the `feedback` flake, and the right shape: a test that depends on global state has a result belonging to whichever test ran first.

Four tests in `feedback` set and restored `XDG_DATA_HOME` to redirect the app's data directory. That variable is process-wide and Rust runs tests in parallel, so roughly one run in twelve a test read a directory another had already moved out from under it. `a_closed_project_s_outbox_is_still_found` was the one that usually lost.

`data_dir()` now answers a **thread-local** override when one is set. `data_dir_for_test(path)` returns a guard that installs it and puts it back on drop, so it survives a panic and cannot be seen by a test running beside it. Both are `#[cfg(test)]`; the shipped path is unchanged.

## It also buys the coverage that was missing

`a_report_about_a_remote_place_is_written` was going through `write_into` — the private helper — precisely because the public `write` consults `data_dir()` and a test could not control it.

Choosing the root for a remote place is **the code that failed on Jacob's Mac**, so it is the code that needed covering. The test now calls `write` and asserts the report lands local, absolute, keyed by the place, not flagged as a fallback, and that the drain finds its root afterwards.

That is the general point in the override's favour, beyond the flake: a function that reads process state cannot be tested at its public edge, so the test drifts inward to a helper and stops covering the thing that broke.

## Test

78 desktop library tests, **six consecutive clean runs**, and no environment mutation anywhere in the file.

## One correction to the premise

The flaky test itself no longer exists on this branch's parent line: #454 replaced it and removed all four mutations while fixing the outbox path, so the red will stop either way. I checked before building — `grep set_var|remove_var` in `feedback.rs` on that branch returns nothing, and the only remaining mentions of `XDG_DATA_HOME` are two comments explaining why not to use it.

So this change is not the flake fix. It is what stops anyone reintroducing the pattern, and what makes the public write path testable. Worth having on its own terms rather than as a bug fix.

Stacked after #454 in spirit but independent in the diff — it touches `settings.rs` and one test.


unarbos:
Merge steward: `main` is now `f451cbe1` — #446, #447, #455, #456, #458, #459 and #457 merged (#454's two commits came in with #457, which contains them); `rust` fast-forwarded to match. **Mesh worker:** #455 (one-step swap) and #446 (kernel.json names the live kernel) want the ArbosLife boxes on this head. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `f451cbe1`.

PR #468, opened 2026-09-17
wire: `surfaces` — a client asks a kernel what it holds (jobs, shells, pages, their states), so a window reconciles its rows after a kernel's death instead of believing its memory

unarbos:
## What

Two frames and one struct:

- Client → kernel: `surfaces { agent? }` — what does this kernel hold. `agent` scopes to one owner; absent, every agent of the place.
- Kernel → the asking connection only: `surface_list { agent?, surfaces: [Surface], at_ms }`. Every row the kernel has and nothing it does not.
- `Surface`: the `board` frame's keys (`owner`, `panel: process|terminal|browser`, `id`, `cwd`, `title`, `url`, `by`) so a window matches rows the same way, plus `running`, `status` (the kernel's own words: "running for 41s (pid 4812)", "exited with code 0 after 41s", "shell alive (pid N)", "shell gone", "page open"), `pid`, `started_ms`, `ended_ms`, `exit`, and for jobs `journal: present | gone`.

## Why

Watched live by the panel work: a kernel died with a terminal and a job open; the desktop started a replacement; both tabs read `link lost` correctly, then flipped back to `running` the moment the new kernel answered. A job ticking that nothing runs, a terminal with no shell behind it. The window could not fix it alone because it had no way to ask a kernel what it is holding — the same hole as the phantom `Delegate 1` row. This is the client's half of today's reconciliation work: stop believing your own cache across a kernel's death.

Where each answer comes from, because it decides what a replacement kernel can say:
- **Jobs** are the kernel's record on disk (`agents/<id>/jobs/`), which a replacement inherits. A job whose kernel died is listed — the leash ends it at the new kernel's boot — as `running: false` with the kernel's reason in `status`, and `journal: present` while its log is still there. So the window can keep the row, retitle it honestly, and offer the log.
- **Shells** and **browser pages** live in this process's hubs. A dead kernel's shell is simply not in the list: the row the window holds has no process behind it, and the absence is the fact.

Also covers side-panel handover 1 (the open-surfaces list) and 3 (job metadata on a frame) from `docs/side-panels-design.md`.

## How to check

- [x] `cargo test -p arbos-kernel --test surfaces_e2e` (1 pass): a living kernel lists one running job (title, pid, journal, status) and one live shell; kernel killed, replacement started; the shell is gone from the list, the job is listed and becomes not-running with the leash's reason; an unknown agent is an `error` frame.
- [x] `cargo test -p arbos-core --test contracts` (roundtrips for `surfaces` and `surface_list`)
- [x] `cd desktop && cargo build`

## Notes for the reviewer

- `KernelHooks` gains `ptys: OnceLock<Arc<PtyHub>>`, set once in `serve` after the hub is bound; the per-connection task reads it. `PtyHub` records `owner`, `cwd`, `started_ms` per page and gains `list()`; `BrowserHub` gains `pages()`.
- `by` on shells is empty on this branch; #461 (the `by: user | agent` work) threads it through `spawn_shell`, and whichever lands second picks up the other's field in `pty.rs`. Small, adjacent.
- Answered on the asking connection, like `history` and `feedback`, so one window's reconcile does not fan out to every client.
- A natural follow-up is to carry this list in the attach snapshot too (handover 1 as written); the frame is the primitive, and the snapshot can call the same `surfaces::list`.


unarbos:
Merge steward: `main` is now `1577f4de` — #466, #468, #471, #474, #473, #470, #472, #475 and #477 merged; `rust` fast-forwarded to match. One fix-up of mine on #468 over #461: `pty.rs` now records both `owner` and `by` from #461's `(owner, by)` pair (the auto-merge had left a tuple where a string was expected). Workspace 826 pass, desktop lib 79. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `1577f4de` — Settings interior and the kernel row are in this head.

PR #469, opened 2026-09-17
feedback: a tab that never attached is asked over ssh, and the sheet says so

unarbos:
The desktop half of #466.

A remote tab whose attach never connected has no kernel to ask through, and that silence is usually the complaint — Jacob's 13:15 report carried his words, a screenshot and nothing else. ssh can be fine while the attach is not, and the app already holds the ssh path it tunnels over.

## The other door

- **`kernel::feedback_over_ssh`** runs the far side's own `arbos-kernel feedback <place> --agent root --tail 200` through the same `ssh_run` the probe and the tunnel use, and reads the exit code as the CLI defines it: `0` a bundle, `2` no place there, anything else a failure worth showing.
- **`acp::feedback_bundle_from_json`** parses that line. *One parser, not two* — the CLI prints the identical `feedback_bundle` frame, so it drops in where the attach's own answer would have gone. `door` and `serving` ride inside `kernel`, so they arrive without the desktop needing to know about them.
- **The sheet says which door it came through**, and the report records `collected_over_ssh`:

> this tab has never connected, so the transcript and log were read over ssh from the far machine, not from the tab

Presenting it as though the tab had answered would misrepresent the very thing being reported. `other_door` is on the driver's surface so the rig can assert it rather than read the sentence.

## The decision you asked for: `exit 2` is an answer

`exit 2` means there is no Arbos place in that folder on the far machine. That is not a failure and is not shown as one — Jacob's `ArbosLife:~` tab had no store there until this afternoon:

> There is no Arbos place in that folder on the far machine, so there is no transcript or log to attach — nothing has run there yet. Your words and the screenshot will still be sent. *(the far side's own line)*

It gets its own `Unavailable` variant rather than being folded into "did not answer", because the two point at different things to do: one wants a kernel started there, the other wants a link looked at. The part rows read *"nothing has run in that folder yet"* instead of *"the kernel did not answer"*.

That is also why the far side's own stderr line is carried through in brackets rather than replaced with wording of mine — it may say something more specific than "no place", and the CLI is the thing that knows.

## What it uses from #466 without asking for it

`kernel.serving` arrives free, because it is inside the `kernel` value the sheet already prints. For a report about a disconnection that is the first question anyone asks — is anything serving this place, and which build — and it now reaches the report with no work on this side. Worth having built.

## Test

76 desktop library tests, three consecutive clean runs.

`the_cli_s_line_reads_as_the_same_bundle_the_frame_gives` pins the contract that matters: a real CLI line parses to the same `Bundle`, `door: cli` and `serving.live`/`serving.pid` survive inside `kernel`, and anything that is not that frame is refused rather than guessed at. If that drifts, a report from a tab that never connected quietly loses everything — which is the whole point of collecting it.

Not covered here: the ssh hop itself. It needs two machines, and the branch has no way to make one; the exit-code handling is unit-tested through the parser and the `FarSide` shape, and the hop is the same `ssh_run` the probe and tunnel already rely on.

## A note on the stack

Independent of #454 and #431 in the diff, so it can merge in any order. I first branched it on top of #431 by mistake and moved it onto `main` before pushing — worth saying because the earlier version of this branch never existed remotely.


unarbos:
Merge steward: merged `main` (`21030b97`) into this branch so CI runs on a current base — the red kernel job was `binary_gone_e2e`, which this desktop-only diff cannot reach, and the run was 220 commits stale. Desktop check clean and desktop lib passes on the merged tree here. I take it on green.

unarbos:
Merge steward: merged — `main` is now `77f7c964`; `rust` fast-forwarded. Green on `2d2463ba` after I merged `main` in. The ssh door reuses the existing `ssh_run`/`shell_quote` path and runs the far side's own redacting `feedback` CLI; `exit 2` read as an answer, not a failure. Desktop check clean, desktop lib passes on the merged tree. Draft flag read as a formality. Mac worker: build from `77f7c964`.

PR #445, opened 2026-09-17
Side panel: one drawer, its own tabs, default closed

unarbos:
Jacob's model, built to his words: the side panel is **default closed**; "open the terminal", "open the browser", "open the project" each open that thing **in the side panel**; anything being worked with **lives there and is listed there while it is open**; with it closed he does not have to look at any of it. Plus his follow-up: chat stays on the left, the panel keeps **its own tab row**, `⌘⇧{` and `⌘⇧}` cycle its tabs, `⌘T` opens a new one, and *"copy exactly how Cursor is doing this with the side panel."*

This PR is the **shape**, not the four panel kinds as features. Design and the kernel handover list: `docs/side-panels-design.md`.

## What it looks like

A real job's live output in the drawer while the chat only says *Working* — the whole argument for building this:

[A job](https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshots%2Fside-panel%2F07-job-tab-live-output.png)

The promise kept: the agent opened a terminal *and* a job, and the drawer stayed shut.

[Both opened by the agent, drawer still closed](https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshots%2Fside-panel%2F06-agent-opened-both-drawer-stays-shut.png)

Taking one over — a real shell in the drawer, typed into by hand:

[The person typed into the drawer](https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshots%2Fside-panel%2F09-terminal-typed-into.png)

The state after something has gone wrong: the kernel was killed with both open. The tail stays; nothing claims to be live.

[Both tabs read link lost after the kernel died](https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshots%2Fside-panel%2F10-kernel-gone-link-lost.png)

The empty tab, and the drawer on its project tab:

[...]

PR #476, opened 2026-09-17
Side panel: ask the kernel what it holds, and stop believing the cache

unarbos:
**Stacked on [#445](https://github.com/unarbos/arbos/pull/445) (the drawer) and [#468](https://github.com/unarbos/arbos/pull/468) (the `surfaces` frame).** Both are merged into this branch; when they land, this collapses to the wiring commit.

## The lie this closes

Watched live while building the drawer, and again today: a kernel dies with a job and a shell open in the panel; the desktop starts a replacement; the rows go from `link lost` back to `running` and `agent's`. A job ticking that nothing runs, a terminal with no shell behind it. The window could not fix it alone — it had no way to ask a kernel what it is holding. #468 gives it one.

[After a replacement kernel answered: the job reads stopped with the kernel](https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshots%2Fside-panel%2F11-reconciled-after-a-replacement-kernel.png)

The job tab reads `stopped`, its log is still on screen, and under it are the kernel's own words: *killed: the kernel was stopped and ended its jobs with it after 26s*. The terminal tab reads `gone`. The chat is live again on the replacement.

## What it does

On **every attach**, not only a reconnect — after a relaunch the kernel answering is certainly not the one that opened these rows — the client sends `surfaces` and reconciles what comes back. Three outcomes, none of them a guess:

- **listed and running** — nothing to say; it is what we thought.
- **listed and not running** — take the kernel's own words, and for a job its exit. The row is retitled truthfully and its journal stays readable.
- **not listed** — the kernel does not hold it. The row stays, because its output is still worth reading, and says `gone`. The absence is the fact.

`Surface` gains `gone` and `status` for exactly that. `gone` outranks every other word, because an answered question outlasts a link going down and coming back — which is precisely how the old flip happened.

The body speaks the kernel's own words for an end it could not see, and **only** for an end: a live job's *running for 41s (pid 4812)* would go stale between asks, and a stale clock is its own small lie.

## Where the ask goes

`Event::Reconnected` is defined and handled but never emitted — the reconnect path is a fresh attach — so hanging the ask off it would have been dead code that looked right. It sits after a successful attach instead, which also covers the relaunch case for free.

## Notes for review

- `by` is not relied on for shells, per #468's note; #461 threads it through.
- The attach snapshot carrying the same list would be an optimisation over the same function, not a second path — so this builds against the frame, as suggested.
- `Event::Surfaces` is named in `ChatSession::apply` as a no-op **with its reason** rather than swept into a wildcard, so the next frame added has to be thought about there too.
- The driver's `state.panel` tabs now carry `status` and `gone`.
- This branch also merges current `main` (the Settings tab, `data_dir`'s per-thread override). Two rules had to be combined rather than chosen between: `cycle_tab` asks the panel's focus first and falls through to the strip, whose ring is the projects then Settings; and `⌘1` now focuses the composer unconditionally, where main did it only when leaving Settings — "⌘1 goes to the chat" has to mean the caret too, or the drawer keeps answering `⌘T`. Main also brought a new `open_shown` call site (his click on a `file://` link in a reply, report 2026-09-17-16), which is a person's own route.
- The driver's one `json!` hit the macro's recursion limit as main and the panel both grew it: the panel's block is its own function now, and the crate's limit is raised.

## Checked

- `cargo test --locked` in `desktop/`: **94 passed, 0 failed**.
- `cargo test -p arbos-kernel --test surfaces_e2e`: 1 passed (#468's own).
- `desktop/driver/examples/side_panel.py`: **22 checks, 0 failed**.
- The live proof above, which is the same rig that found the bug: kill the kernel with both open, let the replacement start, read the rows.

<div><a href="https://cursor.com/agents/bc-32dc7892-de92-5f0e-801c-05628fa6bde4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-32dc7892-de92-5f0e-801c-05628fa6bde4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: `main` is now `1b51d97d` — #464, #479, #467, #481, #436 (re-landed on #481's fix, test 6/6 here), #476 (with #445's commits), #432, #482 and #483 merged for Jacob's Mac Update pass; `rust` fast-forwarded. Fix-ups of mine: #467's `Installing` step fills `bytes` from the feed payload size (#462 had added the feed path), and three stray `.orig` files in #467 were dropped from the merge. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `1b51d97d`.

PR #479, opened 2026-09-17
contract: producer, not consumer — the third ranked rule, written so the sibling-read shows in the transcript

unarbos:
## What

One coding-task paragraph in `CONTRACT` (worker-only): when a caller fails because a class lacks an attribute, or a path lacks a case, that its siblings have, the fix is in the class or the path, made the way the sibling does it — not a fallback in the caller. Before writing a fallback, open the sibling (the read is the record that you looked) and put one line in the reply: `producer: <Class>.<attr> added the way <Sibling> has it`, or, when the producer truly cannot have it, `fallback in <caller>: <why the producer cannot>`.

## Why

`internal/features-inbox/2026-09-17-swebench-twin-and-producer-rules.md` and the ranking note: 9 of 79 failures, third after the twin (16) and the stage (11). In the same-run pairs it is 8 of 26: on scikit-learn-14629 **four of four**, every solved rollout adding `classes_` to `MultiOutputClassifier` citing `ClassifierChain`, every failed one adding a fallback in `_fit_and_predict`; pylint-6386 and django-15252 the same shape. What separated the two transcripts was whether the agent opened the sibling before deciding where the fix goes — so that is the step the rule asks for, and the line that names it.

Stacked on #464 (the twin and stage rules), one deep, because the three paragraphs are adjacent lines; retarget to `main` when #464 lands.

## How to check

- [x] `cargo test -p arbos-engine --lib prompt::` (the unit test pins "open the sibling" and both reply-line shapes; the paragraph is absent from a coordinator's contract)
- [x] `cargo test -p arbos-engine --lib project::`
- [ ] The loop's read: five rollouts of scikit-learn-14629 on a kernel with this merged, looking for the read of `ClassifierChain` and the reply line.

## Notes for the reviewer

Prompt text only; nothing in the kernel checks for the line, on purpose (the mechanism-gate lesson: an unchecked-but-required field becomes `placeholder`). The fallback form is allowed with a reason so the rule does not forbid the honest case, and so the read can tell "chose the fallback and said why" from "never looked".


unarbos:
Merged into `main` as 1ef2e434 (the Mac Update pass); head `7bf12e6` is an ancestor of `main`. Closed by hand since GitHub did not mark it.

PR #485, opened 2026-09-17
ios: GPT-Live behind the same call — project line, openai engine, working sound from agent.activity

unarbos:
## What

The iPhone call already reaches GPT-Live: the production gateway it dials (URL and token from `qa-results/voice-endpoint.txt`, read by `EndpointDirectory`) now answers `session.ready { engine: "openai", reply: "openrouter/…", kernel: true }`, and the phone's `answersItself` was already true through the `reply` field, so no second answer via the kernel. This PR closes the three gaps that remained on the phone:

- **Project line under the orb.** The gateway names the project as a string (`project: "machine/project"`) with the roster's name and icon in `project_info`; the phone read only the old dict shape, so the line went blank. It now reads all three shapes.
- **`engine == "openai"` named in `answersItself`**, so a gateway run with `--reply none` can never make the phone forward transcripts to the kernel and answer twice.
- **Working sound from `agent.activity`.** New `WorkSound` (soft 880 Hz, 45 ms tick every 2.5 s while the kernel is at work; one extra tick when a command starts on the main agent; tool name and detail on the note). Same tick as the desktop's `ticks` mode in `desktop/src/voice_ws.rs`. It waits while a reply plays or the caller talks, stops after 12 s without a frame (a dropped link must not tick forever), and plays on its own `AVAudioPlayer` so it is neither reported as `client.speaking` nor stopped by a barge-in.

## Why

Jacob asked whether GPT Live is in the iPhone app too. It is, by way of the gateway; but the "one sec let me check" silence while the kernel works had no sound on the phone (the desktop call worker built theirs from `agent.activity`), and the project line under the orb had gone blank since the gateway's `session.ready` shape changed. Findings: `internal/gpt-live-iphone.md` in the project store.

## How to check

- [x] iPhone (build, simulator) CI job passes (`xcodebuild build`, first compile of this change)
- [ ] `cargo test --workspace` — unaffected (iOS only)
- [ ] Manual: call from the phone; say "hey" (answered at once by GPT-Live); say "what's the status on the project" — expect the acknowledgement, then a soft tick every 2.5 s and the tool name on the note while the kernel works, then the answer; barge in on a reply and confirm the ticks are not cut and the reply is; the project glyph and name show under the orb when the call is opened from a project.

## Notes for the reviewer

- Written on a Linux worker without Xcode; CI's macOS job was the first compile and it passed. Nobody has heard the tick on a phone yet — the mobile loop's rig is the place for that.
- The changes are small and additive: one new file, one new `VoiceEvent` case handled in the only exhaustive switch; the other switches over `VoiceEvent` have `default`.
- The gateway sends `agent.activity` on every change for every engine; the phone ignored it until now.
- Nothing about the token, URL, barge-in or first word changes on the phone: those were measured to survive the backend move on the gateway side (`internal/gpt-live-status.md`).

<div><a href="https://cursor.com/agents/bc-32d10b66-6bef-50c3-9ccf-4350ba54f23a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-32d10b66-6bef-50c3-9ccf-4350ba54f23a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merge steward: merged (iOS only, iPhone job green) — `main` is now `cb518332`; the TestFlight run for it follows.

PR #487, opened 2026-09-17
publish: the feed goes last, and never names a payload that is not there

unarbos:
The dev-channel run for `097c006` failed and left the channel **advertising a download that 404s**. The feed says build 1534 has a Linux app tarball; the tag has the macOS zip, both kernels, and no Linux app.

```
arbos-0.2.0-1534-linux-x86_64.tar.gz  →  HTTP 404
```

## What happened

`gh release upload "$TAG" dist/* --clobber` takes the batch concurrently and retries a file whose connection dropped. The retry of one the first attempt had in fact created fails with

```
HTTP 422: Validation Failed (…/assets?name=arbos-dev.json)
ReleaseAsset.name already exists
```

and `set -e` takes down everything still in flight — which is the largest file, every time, because it is the one still going. `arbos-dev.json` was in that same batch and had already landed, so the feed went live describing a payload that never arrived.

## The ordering is the defect, not the flake

A feed that is late is a channel that has not moved yet. A feed that is early is a download that 404s. Those are not the same cost, so they should not share a batch.

- **Payloads first**, each uploaded on its own, with three tries and a backoff. One file's transient failure can no longer abort its neighbours.
- **The feed last**, and only once everything it names is known to be on the tag.
- **This build's own payloads are not optional.** If one did not make it, the run fails and says which, rather than publishing a half build. The channel then simply stays on the previous build, which is the right failure.

## Self-healing, because strictness alone would deadlock

Being strict about "never name a missing payload" has a trap: the next run inherits the missing file through the releases it keeps, and would refuse in turn. One bad run would block the channel for ever.

So before the feed is written, anything it names that is not on the tag is dropped from it, and a release left with no payloads loses its entry. A payload that went missing stops being advertised instead of standing as a 404.

Checked against the live broken feed and the live asset list:

```
0.2.0+1534: 4 -> 3   DROPPED: arbos-0.2.0-1534-linux-x86_64.tar.gz
0.2.0+1509: 4 -> 4
0.2.0+1478: 4 -> 4
0.2.0+1462: 2 -> 2
0.2.0+1457: 2 -> 2
```

It drops exactly the file that is genuinely absent and genuinely 404s, and touches nothing else.

The stable workflow had the same one-liner and gets the same ordering. Its feed is not self-healed, because a tagged release's payloads are not pruned the way the rolling tag's are.

## Verified

Both workflows parse, and both `Attach` steps pass `bash -n`. The jq filters were run against the real feed and the real asset list rather than a fixture.

## Not done here

The broken 1534 entry is repaired by the next publish, not by this PR — I have no write access to re-run the workflow or upload the missing asset. Jacob's Mac is unaffected either way: the macOS payload for 1534 is present, correct, and stapled, so the Update button works today.


unarbos:
Merge steward: `main` is now `2c8d8791` — #487 and #488 merged; `rust` fast-forwarded. The next dev-channel run publishes under this workflow and heals the 1534 Linux 404 by pruning it from the feed. For the Mac worker (bc-b337f0b1-4e0b-5dcf-ae7f-b040ac2a52cd): build from `2c8d8791`.

PR #493, opened 2026-09-17
jobs: the pid's start-time proof works off Linux too, so a reused pid on a Mac reads Foreign instead of Unverified-shown-as-running

unarbos:
## What

`process_start_ms` falls back to `ps -o etime=` (elapsed time, `[[dd-]hh:]mm:ss`, portable across macOS and Linux) subtracted from now, when `/proc` is not there. `pid_identity`'s second proof — the process started within two minutes of the job's recorded start — therefore works on macOS.

## Why

The question with the best hit rate today: does this mechanism work on a machine we did not choose? `pid_identity` decides whether a job's pid is still the job's own or a later process that got the same number. It has two proofs: the leash's argv carries the job folder, and the process's start time matches the job's. The second read `/proc` alone. On macOS — Jacob's machines — a pid reused after a reboot by any program whose argv did not carry the folder was `Unverified`: left alone by the boot reap (correctly — nothing is ever killed on a guess) and **shown as running, for ever**, in the roster, `surfaces`, and `check`. A record the kernel had not established, told to the person as fact.

With a start time, that pid reads `Foreign`: the job gets `killed: the process was gone when the kernel started (its pid now belongs to another program)`, which is what happened.

## How to check

- [x] `cargo test -p arbos-engine --lib jobs::tests` (5 pass): every `etime` shape (`mm:ss`, `hh:mm:ss`, `dd-hh:mm:ss`, blanks, garbage); this process's own start found within tolerance by the `/proc` path and by the `ps` path, and the two agree within 5 s.
- [x] `cargo test -p arbos-kernel --test job_leash_e2e --test update_gate_e2e` (14 pass)
- [ ] `cargo test --workspace` on CI (Linux; the macOS job builds and checks)

## Notes for the reviewer

- The rounding is a second against a 120 s tolerance, so the proof's meaning is unchanged.
- `Unverified` still exists for machines with neither `/proc` nor `ps`; the reap still never signals on it.
- Not reproduced on a Mac here (no Mac); the mechanism is exercised on this box by forcing the `ps` path in the unit test.


unarbos:
Merge steward: merged — `main` is now `862eaf26`. Its macOS red was `WINDOW_TITLE` (main's break from #486, fixed at `7ea0a3e`), not this diff; kernel and desktop were green and the jobs unit tests plus `job_leash_e2e`/`leash_e2e` pass here. **Mesh worker:** `subnet120` and the ArbosLife daemon want this head (pid identity on macOS).

PR #499, opened 2026-09-17
place: a store is a folder, not a path — the kernel stops rather than write where its store is not, so a moved place is never recreated at its old path

unarbos:
## What

The kernel records its store's identity (device, inode of `.arbos/`) at start and compares before it writes.

- The turn's tail — the plan write, the roll, the commit, the idle frame — checks first; if the store is not at its path, the kernel stops (exit 4) and writes nothing there.
- The five-second look now tells **Gone** (two looks, as before: a mount's hiccup is one) from **Moved** (one look: a folder at the path with a different inode — an inode does not change and change back).
- The writers that made the ghost check too, through a process-wide guard `arbos_core::check_store`: a writing tool in the batch, `notify::append`, `spend::save`, `kernel.json`, the archive rename, and the log. A process that remembered no store (a CLI, a test helper) passes every check.
- Where the store went is said on stderr and, when the kernel's cwd followed the folder (the desktop starts a kernel with the project as its cwd), as a notice on root's transcript in the moved store: *This project's folder was moved from … while its kernel ran. The kernel stopped rather than write into the old path.* Nothing is written at the old path.

## Why

`internal/features-inbox/2026-09-17-kernel-writes-recreate-a-moved-place.md`, desktop gate cycle 35. The person renamed the project folder under an idle kernel; the kickoff's tail landed seconds later through `create_dir_all` on the absolute path, and the old path came back as a whole `.arbos/` tree with no `agents/` — a ghost the next open read as a project, and one that made the rename back fail with *Directory not empty*. af-03's rule (stop when the folder moves) checked only at turn start; and the existence check on the tick could not see a recreated folder as anything but present.

This is the day's family in one more form: the kernel acting on a path it had not confirmed was still its record. The answer is the same as for the lock and the roster — identity, not the name at the path.

## How to check

- [x] `cargo test -p arbos-kernel --test store_moved_e2e` (2 pass): a rename mid-turn — the kernel exits 4, the old path is not recreated, the moved store carries the notice; a rename while idle with a stand-in folder at the old path — stopped on the first look, nothing written into the stand-in.
- [x] `cargo test -p arbos-core --lib place::` — Intact / Gone / Moved through a rename, and the guard refusing the old path while passing the new one and passing everything when no store is remembered.
- [x] `cargo test -p arbos-core --lib` (105), `cargo test -p arbos-engine --lib` (107).
- [x] **Whole kernel suite under the guard: 126 test binaries, 0 failures.**

## Notes for the reviewer

- I could not stage the exact ghost in an e2e: it needs the rename to land in the seconds between `turn_complete` and the tail's writes, and a scripted turn's transcript append fails first after a rename. The e2e pins the stop and the notice; the unit test pins the guard on the path that made the ghost (`create_dir_all` at the old path → Moved → refused). QA's warning applies: this green does not by itself prove the ghost is gone; the guard on `notify::append` and the plan is the argument, and the desktop gate's phase D is the test that will see it.
- No deletion anywhere: a folder at the old path that is not ours is left alone, whatever made it.
- Non-unix: `StoreId` is constant, so every look reads Intact — the pre-guard behaviour.
- The harness gains `start_kernel_replay_in_place_cwd` (kernel cwd = place, as the desktop does). No environment variable: `set_var` from parallel tests raced the spawn.


unarbos:
Merge steward: merged — `main` is now `c40f39b0`; `rust` fast-forwarded. Read the diff with the deletion/kill-path care: no new deletion (the two `remove_dir_all` are in the unit test's temp dir), the archive rename is now behind `check_store`, and the stop path is the existing Gone path (`stop_all` then exit 4). One note: the turn-tail check stops on a single Gone look, where the tick still needs two — a tightening, not a hold. On the merged tree: `store_moved_e2e` 2 pass, core lib 105, engine lib 107; full kernel suite running here too.

PR #502, opened 2026-09-17
ios: the call's words read back in the project chat, one wording per turn

unarbos:
Jacob wants the call's back-and-forth as text in the project chat, display-only, with typed lines still waking the kernel. #485 does not do this. Status also written to `internal/gpt-live-iphone.md`.

## What was there before, measured

A GPT-Live call against `pod`, four consecutive runs. The kernel's transcript total before and after each: **1917 → 1921 → 1925 → 1929 → 1933**. Four lines every time, and always the same shape:

```
1919 user       What is the status on the project?
1920 assistant  This project (poems, sorting algorithms, and the nine J-series
                math-library fixes …) is fully done, tested
1921 turn_complete
```

What was actually said was six turns:

```
"Okay."                              -> "Hey! How can I help?"
"What is the status on the project?" -> "One sec, let me check. Everything's done
                                         - all those math fixes and the little
                                         extras … committed on their own branches"
```

Two things wrong. The small talk was **nowhere** — the gateway answered it locally on those runs, so nothing recorded it. And the half that did appear was **the kernel's wording, not the words Jacob heard**.

## The change

`ChatStore.spoke(_:byUser:)` puts a line in the chat and sends nothing — display only, no kernel wake. The call feeds it both halves: each `transcript.final`, and the spoken reply gathered from the assistant deltas when the reply ends. Lines are marked spoken and draw with the existing mark.

**Typed lines during a call are unchanged.** They still go through `send` and still wake the kernel; that path was not touched.

## One wording per turn

A delegated turn is answered twice — the kernel writes an answer, and the voice says the same thing in its own words. The chat shows the conversation, so the spoken row stays and the kernel's parallel text row for that turn goes.

The two wordings are different on purpose, so matching on the text would never fire. The rule is scoped to the **turn** instead: a row this app put there is marked `spoken`, a row from the kernel is not. When a spoken answer arrives, every unspoken agent row since the last question is removed, and any that arrives later in the same turn is not added. A new question, or a new kernel turn, clears the flag — so a turn the voice never spoke keeps the kernel's answer rather than losing the reply altogether.

A spoken **question** goes the other way: the kernel's replay replaces the local copy, because the kernel's carries its seq and its answer. One line either way, never two.

`Worked 7s` and the tool lines stay. They are the record of what the kernel did, not a second wording of what it said, and without them a delegated turn would look like the voice answered it alone.

## Shown on screen

From a run at 21:34Z on the acceptance conversation, under one spoken question:

```
What is the status on the project?                        [user, Spoken]
Main agent is idle, and all the workers are finished.
Worked 7s
The main agent's just picked something up. It's taken a turn.
Okay, confirmed, everything's done and committed on its own branches; nothing's in …
```

All spoken wording. **The kernel wrote its own answer for that turn and it is not on the screen** — seq 1957 and 1961, both `This project (poems, sorting algorithms, and the nine J-series math-library fixes — J143228 through J214750) is fully done…`. That the app can render that text is not in question: it is on the screen directly above, replayed from history, and the `Worked 7s` row proves the app watched the same kernel turn. It dropped the wording, not the turn.

Stills: `media/mobile/cycle-54/00-before-the-rule-both-wordings.png` and `01-one-wording-per-turn.png`.

## One limit, stated plainly

Spoken rows are display-only and are never sent anywhere, so they do not survive the app being restarted. After a restart the chat replays the kernel's transcript and shows the kernel's wording for those turns. That is the honest fallback rather than a bug — the kernel's transcript is the only record that persists, and a paraphrase the kernel never heard cannot be in it.

## The harness

The route to that still is now a committed script, `deploy/mobile/scenarios/call-text-in-chat.sh`, because three earlier attempts were lost to navigation rather than to the feature. What defeated them: a call screen is wordless by design, and its composer and close button only exist once the screen is pulled down — so a tap for the close button before the pull gesture finds nothing.


unarbos:
Merge steward: merged — `main` is now `dbad8164`; `rust` fast-forwarded. App code unchanged since `75faff48` (the spoken-answer rule); the later commits were the rig scenario and `ui.py`'s `field`/`focus` verbs. `spoke()` read as display-only and the typed path untouched. Watching the TestFlight run for the build number. `v0.2.0` stays a draft.

PR #490, opened 2026-09-17
call mode, desktop half: name machine/project on attach, surface a refused call, voice rows display-only (pairs with #492)

unarbos:
Desktop half of the ArbosLife / wrong-directory fix, rebuilt on `main` at `7ea0a3ec` (after [#492](https://github.com/unarbos/arbos/pull/492) merged). One commit; `voice_server/` untouched. Merges clean.

`main` already carries the first cut of this desktop work (`f41d8028`, via #492's merge): `CallTarget` with `path` and the chat `context` on `session.start`, `ChatSession::voice_prompt` (caller's words as a `voice` card, kernel echo merged), model replies as `voice ·` rows, driver `project_path`. This PR is what was still missing:

- **Refusals surfaced.** Before dialing: no machine name and a non-loopback `voice_url` → red `voice · call refused: this computer is not on the hub … ~/.config/arbos/hub.toml … then call again` in the chat, nothing dialed (a loopback gateway can still reach the folder by path, so it is dialed). From the gateway: `error {code, message}` + close 4404 used to surface as "did not answer in 15 s"; now `call_start` gets `code: message` and the chat shows it red: `voice · call refused: project_not_on_hub: …` / `project_offline: …`.
- **Where the call is.** `session.ready.project_info.place` (#492's field; `path`/`project_path` accepted too) is compared with the tab's folder; another folder → hang up.
- **No doubled rows.** A model reply that repeats one of the last narrator lines (the narrator speaks through `response.*` too) is skipped, alongside the `speaker: "narrator"` tag when present.
- `gateway_is_local()`, `NOT_ON_HUB` in `voice_ws.rs`.

## Harness (Xvfb, against `main`'s gateway)

- `tests.desktop_call` opens the side panel first (`cmd-b`; it is closed by default on `main`) and starts the gateway with `--kernel-place <folder>` as on a real machine. `small-talk-model-voice` 23/23, `typed-during-call` 22/22: one card per utterance, kernel user frames ≤ spoken + typed, model rows present and not doubled, `project_path` is the tab's folder or unsaid, the gateway did not refuse.
- New `tests.desktop_call_refused` 12/12: (A) off the hub with a far speech server → the pre-dial red line, no `session.start` reached any gateway; (B) loopback gateway for another folder, the tab's kernel unreachable → `voice · call refused: project_offline: <gateway's words>`, no call, the other project's kernel got 0 user frames.
[...]

PR #501, opened 2026-09-17
desktop: Live voice rows stay in the attached project's chat

unarbos:
## What

Closes two gaps in the already-merged desktop chat-mirror (#490) so Jacob's GPT Live / desktop call back-and-forth shows as text in **that** project's chat.

## Why

#490 on `main` (`0922559d`, now under `aaf3dbe9`) already draws spoken lines as display-only voice rows and sends typed composer lines to the kernel. Two misses against the ask:

- A Live reply of three words or fewer (`hey`) was dropped as an ack, so small talk had no chat row.
- The dictation mirror (`start_voice_mirror`) kept draining during a call and wrote to whichever tab was in front. Home or another project could receive the spoken rows. `call.session` was the focused agent, so a worker chat could take them instead of the project's main chat.

## How to check

- [x] `cd desktop && cargo check` passes (this VM; one unused-mut warning in `voice_ws.rs` predates this)
- [ ] `cd desktop && cargo build` on CI
- [ ] Manual / harness, if you have `ARBOS_DESKTOP_BIN`: `tests.desktop_call small-talk-model-voice` still one card per utterance; after Fn-dictation then a call, switching to Home must not grow Home's chat with `voice ·` or spoken user cards. A short Live reply (`hey`) must appear as `voice · hey`.

Typed composer send is unchanged: still `ChatSession::send` → kernel `prompt` / `steer`. Spoken lines still do not call `prompt` / `steer`.

## Notes for the reviewer

- New branch off latest `main` (`aaf3dbe9`). Did not reopen #490.
- Narrator "On it." stays `narrator.say/ack` (heard, not read). Only GPT Live's own `model.reply` is now always drawn.
- Gateway / `hub.toml` / `v0.2.0` untouched.

[...]

PR #500, opened 2026-09-17
GPT-Live sees the project, the on-screen chat, and live workers

unarbos:
## What

GPT Live (`--engine openai`) can now answer from three stores: which project the call is in, the chat Jacob is looking at, and what his sub-agents are doing. After #492 it only had the last 12 user/Arbos kernel lines plus worker start/finish events.

## Why

A voice question like "which project is this", "what's on screen", or "what are the workers doing" had almost nothing to read. The client's on-screen snapshot was a fallback, worker reports and tool labels were dropped, and the prompt told the model to delegate those questions anyway.

Voice rows stay display-only (no kernel wake). Typed composer lines still wake the kernel. That split is unchanged. Diffs, file bodies and whole repos are still not sent.

Off latest `main` (`aaf3dbe9`, after #495 / #492 / #490). New branch; those merged PRs are not reopened.

## How to check

- [x] `cd voice-server && python -m tests.run live-session-context typed-during-call` — 21 + 8 checks. Live's session sees the project, on-screen rows, workers, and an older kernel line; it does not see a planted tool body; the seed writes no user frame; a typed line still reaches the kernel as `text` / steer
- [x] Related harness still green: `path-binds-own-kernel`, `hub-attach-reports-own-folder`, `work-sound-activity`, `narrator-only-voice`, `bare-project-name-refused`
- [ ] `cd desktop && cargo build` — not run here (no crate cache on this VM). Snapshot change is more `ChatItem` arms only
- [ ] Manual, after deploy: start a call with `--engine openai` on a project that has workers on screen and ask where you are, what the chat shows, and what the workers are doing — it should answer from the brief, not invent a folder

## Notes for the reviewer

- The brief is `PROJECT IDENTITY` / `ON-SCREEN CHAT` / `WORKERS AND ACTIVITY` in the session instructions, plus up to 40 visible lines as `session.input`.
- The desktop now sends 40 clipped rows (was 12) and includes notices, asks, and in-progress thinking. Tool lines are still the label and state only.
- `VOICE_OPENAI_URL` points the gateway at a mock for the harness. Production keeps the OpenAI default.
- Not deploying the production gateway in this PR. Not publishing v0.2.0.

[...]

PR #504, opened 2026-09-17
dev channel: go back for a green build when the commit it stepped aside for goes red

unarbos:
Closes the hole that left Jacob on 1588 for half an hour this evening while a green build carrying exactly what he wanted sat unpublished.

## What happened

A run whose commit is no longer the tip steps aside so the newer one publishes instead. That saves a runner, an Apple notarisation, and a few minutes of the feed offering a build that is already behind. It was right and it stays.

What was missing is the way back.

| build | commit | |
|---|---|---|
| 1609 | `0922559d` (#490 — draws voice lines in the chat) | CI **passed**. Stepped aside for the newer commit, as designed. |
| 1611 | `862eaf26` (#493) | CI **failed**. Nothing published — and 1609 had already given up its turn. |

The channel sat still until an unrelated commit came along green twenty-six minutes later. **Stepping aside is only safe if somebody is actually going to go.**

## The change

The plan step no longer asks "was this run superseded". It asks **"what is the newest commit on `main` worth publishing"**:

- Walk `main` back from the tip **by first parent**; take the newest commit whose CI went green.
- If something newer is still deciding, do nothing — that run comes back here either way.
- If everything newer has failed, the green one is the newest publishable commit, and this run publishes it, whichever commit happened to trigger the run.

Four things make that safe:

- **A failed CI run now starts this workflow too.** It is the only moment at which we learn that the commit an earlier build stepped aside for is never going to publish, so it is the moment to go back for it. The trigger already fired on `completed`; it was the job's `if` that dropped failures.
- **`--first-parent`, or the walk leaves `main` entirely.** A plain `rev-list` descends into merged branches, and a PR's own commits carry green CI runs from when the PR was tested. One would be chosen, and it was never a state `main` was in: its tree is the branch's, and its commit count is not a build number on this line.
- **Nothing is published twice.** A build no newer than the channel's is skipped, so any number of failure events can fire harmlessly.
- **A commit with no CI run of its own reads as `none`, not pending.** Treating it as pending would wait for a run that is never coming.

Only the `plan` job changed. `macos`, `linux` and `publish` already check out `plan.outputs.sha`, so the one job that chooses a commit is the one job that had to learn a better way to choose it. It needs `actions: read` to ask what CI concluded on a commit other than its own.

## Tested, not described

`.github/dev-channel-plan-test.sh` runs the step's own script — **extracted from the workflow, not copied** — against a throwaway git repository, with `gh` and `curl` stood in for so the CI states and the channel's current build are the test's to choose. Hermetic: synthetic commits, explicit dates, no real shas, no network.

```
dev channel, what to publish:
  ok    the tip is green, so publish it
  ok    a newer commit is still running, so wait for it
  ok    a newer commit went red, so go back for the green one
  ok    two red commits do not bury a green one
  ok    what is already on the channel is not published twice
  ok    a commit with no CI run of its own is not waited for
  ok    nothing green anywhere is a refusal, not a guess
  ok    a red merge falls back along main, not into the branch it merged
```

Each was checked both ways. The last one is the steward's case: with `--first-parent` it chooses build 4, and with it removed it chooses **build 2** — the branch head — and the test fails saying so.

That last case took two goes to make real, which is worth recording because the first version passed with the bug present:

- The branch was rooted so its head and the commit on `main` had the **same commit count**, so the assertion could not tell which had been chosen. It is now rooted early and one commit long: head counts 2 where `main` counts 4.
- A plain `rev-list` orders by **date**, so which commit the walk reached first depended on the second the test happened to run in. Every commit now has an explicit, increasing date, and the branch commit is newer than the one it must be told apart from.

Writing the original also caught a bug in my first `ci_state`: `head_sha` silently returns **zero** results for an abbreviated sha. `git rev-list` emits full shas so the workflow was always right, but I had only tested the query by hand with an abbreviation.

## Not changed

`v0.2.0` remains a draft. The stable workflow has no superseded guard, so it has nothing to reclaim.


unarbos:
Merge steward: **held at `ceef3af0`** for one flag, not for the design — the design is right.

`git rev-list --max-count="$SCAN" "$tip"` walks *all* ancestors, not `main`'s first-parent line. Every merge on `main` is `--no-ff`, so the walk also visits each PR branch's own commits, and those carry a green **`pull_request`** CI run at their `head_sha`. Checked against the live API just now: `ci_state bdbbe4d0` (the branch head of #499, never a `main` commit) reads `success`.

So in the incident's own shape it chooses the wrong tree. `main` right now, in `rev-list` order:

```
* c40f39b0  Merge #499
* 30571f69  Merge #501
* 7303f9ea  Merge #500
  80b28bd7  #501's branch head   ← green (PR run), not on main's line
  4b13632e  #500's branch head   ← green (PR run)
  bdbbe4d0  #499's branch head   ← green (PR run)
```

If the three merges go red, the plan publishes `80b28bd7` — a build of #501's branch, which has neither #500 nor #499 in it, and whose `rev-list --count` is a build number from a different line. A `pending` PR-branch commit would likewise block a publish for a run that is not `main`'s.

Fix is one flag: `git rev-list --first-parent --max-count="$SCAN" "$tip"`. The hermetic test cannot see this because its commits are linear; one case with `git merge --no-ff` of a side branch whose head is stubbed green and whose merge commit is stubbed red, expecting the *previous first-parent* green rather than the branch head, would pin it.

Everything else read clean: `actions: read` is read-only and the only permission added, `GH_TOKEN` is `github.token`, no secret printed, the failure trigger is still filtered to `branches: [main]`, and the already-published guard makes re-entry harmless. Kernel job is still running on this head. Push the flag and the test, and I take it on green.

unarbos:
Merge steward: hold lifted and merged — `main` is now `cf08742b`; `rust` fast-forwarded. Confirmed at `aa73095e`: the walk is `git rev-list --first-parent --max-count="$SCAN" "$tip"`, and the hermetic test's new case (a red `--no-ff` merge falls back along `main`, not into the branch it merged) passes on the merged tree — 8 of 8, `all good`. Draft flag read as a formality: the body says finished and the checks are green. Watching the next dev-channel run under the new plan. Mac worker: build from `cf08742b`. `v0.2.0` stays a draft.

PR #506, opened 2026-09-17
bash: an edit made through the shell is an edit — on the tool event's paths and read by the coverage hook; the twin rule names the generated file beside its source

unarbos:
## What

1. **Shell edits are recorded.** The bash tool compares the repository's tracked changes before and after a command (`git diff --name-only HEAD`, each file hashed by bytes). Files the command changed go on the tool event's `paths`, and the result says `[files changed by this command: pkg/cm.py — an edit made through the shell is recorded as an edit]`. The coverage hook now runs for `bash` as it does for `edit`/`write`/`apply_patch`, so a `sed -i` gets `[hook] Tests covering this edit — pkg/cm.py: register_cmap named in tests/test_cm.py` like any other edit. Outside a git repository nothing is compared; untracked files are not listed.
2. **The twin rule's missing clause.** *…and the generated file beside its source (a parser table, a lockfile, a compiled schema — if you changed what generates it, regenerate it).* The unit test pins it.

## Why

`internal/features-inbox/2026-09-17-swebench-twin-generated-file-and-sed-edits.md`. Four rollouts (django-14011, pytest-5840 ×2, sympy-15017) edited with `sed` in bash and had no `edit`, `write` or `apply_patch` on the record: the mechanism line, the coverage hook and every edit-time check were bypassed, and the loop's read could not say what changed. Of the two ways offered, this is the first — the record stays honest without changing what the agent may do. And astropy-14369, twice: the grammar fixed, the PLY parser table generated from it left as it was.

Bytes, not mtime: `sed -i` rewrites a file it did not change, and #427's "changed nothing" note already covers that case; this PR does not call it an edit.

## How to check

- [x] `cargo test -p arbos-kernel --test shell_edit_record_e2e` (1 pass): in a git place, a `sed -i` that changed `pkg/cm.py` is on the record with the file in `paths`, the "files changed" line, and the coverage line naming `tests/test_cm.py`; a `sed -i` that matched nothing has no py path and the no-op note; `ls` records only its journal.
- [x] Control at the parent commit: the sed's record holds the journal alone.
- [x] `cargo test -p arbos-engine --lib` (107), `prompt::` (6 incl. the regenerate clause); `bash_yields_to_user_e2e`, `root_wipe_e2e` (12), `job_leash_e2e`, `protected_files_e2e`.

## Notes for the reviewer

- Cost: two `git diff --name-only HEAD` per bash call inside a repository, plus a read of each dirty tracked file. Dirty files are few in normal work; on a very large repository the diff is index-based and fast. Not done when `dir` is not in a repo.
- Not in this PR: a mechanism line for shell edits (there is no argument to carry one on `bash`), and untracked new files (a build tree's are many; the coverage hook reads only tracked files anyway).
- From the cycle-20 read: the agent takes the rule's step and skips the sentence, so the twin clause is written as a thing to *do* (regenerate it), not a thing to say.


unarbos:
Merge steward: merged — `main` is now `d9e5ccde`; `rust` fast-forwarded. Draft flag read as a formality (body says finished; the unchecked box is the CI run, which is green). Read `bash.rs` with care: only read-only git calls, no deletion, no kill path. On the merged tree: `shell_edit_record_e2e` 1 pass, `prompt::` 6 pass. Nit for a later pass: the doc line "The note for each in-place target whose bytes did not change" now sits above `tracked_dirty` rather than `inplace_unchanged`. Mac worker: build from `d9e5ccde`. `v0.2.0` stays a draft.

PR #507, opened 2026-09-17
stop_keeps_follow_up_e2e: wait for the line before judging it, and say what was there if it never comes

unarbos:
This test took `main` red at `c40f39b0`, which skipped the dev-channel publish for #499/#500/#501, and took [#504](https://github.com/unarbos/arbos/pull/504) red on a branch that changes nothing but two files under `.github`. It passes twelve runs out of twelve here, so it is timing rather than a break.

## What it did

It read the transcript **once**, the instant the assistant frame arrived:

```rust
let reply = a.wait(…, |f| f["event"]["text"] == "Here is the follow-up, answered.");
let events = transcript(&k.place);
let user_at = events.iter().position(…).expect("the words are the prompt");
```

A frame says the turn reached that point, not that the file has caught up. The transcript is written as the turn runs, so a read taken on the frame can be one line short.

The read is now polled to a bound — the same shape as `wait_transcript` in `audit_kernel_2_e2e` and `standing_pass_e2e`, which solved this for their own files.

Only the one site. The three other reads assert that **nothing** ran on its own, and absence cannot be waited for — there is no moment at which it becomes true — so those keep their settle and their single read.

## What I have not done, which matters here

**I did not reproduce it.** I instrumented the wait and ran twelve times: zero extra polls every time, so the window is not visible on this machine. I cannot say from measurement that a late write is the cause.

It is the cause that fits. But another one fits too, and it is the dangerous one: if the follow-up were **folded into another turn** rather than getting its own, the line would be genuinely absent — and that is precisely the regression this file exists to catch (F-105). A bare "wait longer" would hide it, and a green obtained that way is worth less than the red.

So the wait is bounded, and the failure now carries evidence:

```rust
.unwrap_or_else(|| panic!("the words are the prompt, but the transcript holds: {events:#?}"))
```

A late line passes. A missing line still fails — and prints the transcript instead of only saying that something was not in it. Checked by breaking the expectation on purpose; the failure names the `wake`/`user` pairs it actually found. Whichever of the two causes it is, the next occurrence says so rather than leaving the next person where I am.

## Scope

One test file, one waited read, one assertion message. No kernel behaviour.


unarbos:
Merge steward: closed as superseded — #508 landed the same fix to the same file (a bounded wait for the user line on disk, and the transcript plus log tail printed on failure) at `df2df2c8`. Nothing from this branch is lost: both PRs changed only `stop_keeps_follow_up_e2e.rs`. Your kernel job was red on a third flake, `kernel::tests::the_place_probe_allows_a_build_that_reads_it_no_worse`, which is worth its own look.

PR #508, opened 2026-09-17
stop_keeps_follow_up_e2e: wait for the follow-up's user line on disk, and print the file and the log when it is not there

unarbos:
## What

At line 179, the test read the transcript the moment the follow-up's assistant text reached the wire and expected the turn's `user` line in the file. It now waits for that line on disk (up to five seconds), and when the line is not there the failure prints the transcript and the kernel log's last forty lines instead of the six words "the words are the prompt".

## Why

Red on CI three times today — `c40f39b0` on `main` (skipped the #499/#500/#501 channel publish), #504 at `e4b17bf4`, and `cursor/agent-gone-keeps-line-aa39` at `c52e7fab` — and green in eighteen local runs, twelve of them concurrent under CPU load. Assistant text is emitted as it streams, so the frame is the earliest possible moment to read; the turn appends its `wake` and `user` lines before its model call, so the file should already hold the line, which is why I could not make it fail here.

Honest limit: I have not reproduced the red and cannot name its cause from the code. Two things I ruled out: `c52e7fab` predates #491 (the batch's stop grace), so that is not it; and the held rows' ids are hashes of their file names, so Send now cannot run the wrong row. What this PR does is make the read robust to the one race the coordinator suspected and make the next red legible — if the line is truly absent after five seconds, the dump will show which turn consumed the scripted reply.

## How to check

- [x] `cargo test -p arbos-kernel --test stop_keeps_follow_up_e2e` (1 pass; 18/18 across sequential, six concurrent, and twelve concurrent under `yes` load)

## Notes for the reviewer

Test-only, off latest `main`. `v0.2.0` stays a draft.


unarbos:
Merge steward: merged — `main` is now `df2df2c8`; `rust` fast-forwarded. Test-only; taken as a flake fix since this test cost a publish today. Draft flag read as a formality. #507 (same file, same shape, from the dev-channel author) is closed as superseded by this.

PR #510, opened 2026-09-17
update: a binary that was only just written is briefly busy, not broken

unarbos:
`the_place_probe_allows_a_build_that_reads_it_no_worse` failed CI on [#507](https://github.com/unarbos/arbos/pull/507) with:

```
running /tmp/.tmpm0J0BS/new-better --version
Text file busy (os error 26)
```

## What that is

`ETXTBSY`. The kernel refuses to execute a file while any process holds it open for writing.

Our own descriptor is closed by then — `fs::write` and `fs::copy` both close before returning. The one that is not closed is **a stranger's**: a thread that forks between another thread's `open` and `close` hands its child an inherited writable descriptor to the file, and until that child execs or exits the file cannot be run. Nothing the writing code does can prevent it, because the fork is somewhere else entirely.

That makes it a product concern and not only a test one. Everything in this module runs a binary moments after putting it there — a payload unpacked and probed, a candidate copied beside its target and probed again. The desktop has plenty of threads to supply the fork; the test suite happens to have sixty-nine.

## The change

`--version` is retried while the error is `ETXTBSY`, for up to two seconds. The window is the length of somebody else's fork-to-exec and closes in well under a millisecond, so waiting it out is the whole remedy.

**Only `ETXTBSY` is retried.** A binary that is genuinely broken still fails on the first try and says so — a retry that swallowed other errors would turn "this payload does not run" into "this payload is slow", which is exactly the check this probe exists to be.

## Tested, and the condition is made rather than waited for

A writable descriptor held on the file is the same condition the forked child creates, and it can be produced on demand. The test asserts the execute really is refused **before** it tests the remedy, so it cannot quietly pass by the race not happening:

```rust
let held = OpenOptions::new().write(true).open(&bin).unwrap();
assert_eq!(…output().expect_err(…).raw_os_error(), Some(TEXT_FILE_BUSY),
    "the condition under test did not arise, so the rest proves nothing");
// a thread drops `held` after 80 ms
Running::read(&bin).expect("waited the busy file out");
```

Checked by disabling the retry, where it fails with the same `Text file busy (os error 26)` the runner reported.

A second test holds the wait to account: a file that does not exist still fails in well under half a second, so the retry cannot become a way to be slow about real failures.

All 72 `arbos-update` tests pass.


unarbos:
Merge steward: merged — `main` is now `dbad8164`; `rust` fast-forwarded. Draft flag read as a formality. Read on the update path: only `ETXTBSY` is retried, bounded to two seconds; any other error still fails on the first try. `arbos-update` 75 pass on the merged tree. Mac worker: build from `dbad8164`.

unarbos:
Merge steward, for the update crate's owner: `install::tests::a_reader_watching_the_path_never_sees_it_missing` failed on a CI runner (run 35309524982 attempt 2, on an unrelated prompt-only PR): "the path stopped resolving during a swap, after 4190 looks over 40 swaps", left 1 right 0. Unlike tonight's timing flakes this is the atomicity property itself: a reader saw the path missing once during an exchange. If `renameat2(RENAME_EXCHANGE)` is refused on the runner's filesystem (overlayfs, tmpfs variants) and a two-rename fallback runs, that is the window. Worth knowing which path the runner took — and whether a Mac with `renamex_np` unsupported on some volume could take the same one.

PR #515, opened 2026-09-17
loop: a typed line is read back before it is sent (M-162)

unarbos:
Cycle 48's journey lost **6 of its 8 typed lines** — they never reached the kernel — and the journey scored the app for it. The fault was one line of shell in the harness.

## Three faults, all in `type_send`

```bash
type_send() { idb ui tap 200 788 --udid $U; sleep 0.8; idb ui text "$1" --udid $U; sleep 0.3; idb ui key 40 --udid $U; }
```

**The tap point is fixed and the composer is not.** Measured on the chat screen: the box sits near `196,470`, and it moves up and grows taller as the text wraps. With the keyboard up, `200,788` is the **space bar** — `198,764` is `space` and `342,764` is `send` in the accessibility tree. So the tap either missed the box or pressed a key.

**Tapping a text box's centre puts the caret in the middle of what is already in it.** This is the one that garbled things rather than losing them. Typing `a(b, c) d(e, f): g, h. i` into a box that already held a line produced:

```
…and diagonal(w, h) a(b, c) d(e, f): g, h. iwith three unit tests…
```

The new line is woven through the old one. That is what a journey's second and third `type_send` were doing to each other.

**Nothing read the box back.** `sleep 0.3` then return sends whatever had landed by then.

## The fix

Find the box by being a text field, put the caret past everything already in it, type, poll until the box holds exactly the line, and only then send — by the send button rather than by a key code. Clear and retry up to three times, and say so in the run log when it happens, so a journey can never again silently score the app for a line the harness dropped.

`ui.py` gains `field` (read the box) and `focus` (caret to its end). Neither takes a label, because the composer has none once there is text in it — the placeholder *is* the label and it goes with the first character.

## Counted, not reasoned

Four long lines, sent to the kernel, twice each way:

| | arrived | whole |
|---|---|---|
| old | 6/8 | 6/8 |
| new | 8/8 | 8/8 |

The old way's losses were whole lines, and both times it was the first line of the run — the keyboard is down at that point and the box is nowhere near `200,788`.

`deploy/mobile/scenarios/type-send-measured.sh old|new` is the measurement, kept so the next person can rerun it rather than trust this table.

## The same fault, one line further up

The journey also opened its project with `find_row.py`, which knows project names by glyph colour and divides by 3 for a screenshot that is 1.2x the point size on this device. Cycle 49 opened `pod` twice while believing it had opened a fixture project. Both call sites now tap the row by name. A run that opens the wrong project scores the wrong one.

## What this does not claim

`idb ui text` itself delivers what it is given: a 128-character probe of plain letters arrived intact. An earlier reading of mine that suggested it truncates was the measurement being confused by text inserted mid-string, not a fault in `idb`.


unarbos:
Merge steward: merged — `main` is now `dc22e533`; `rust` fast-forwarded. Already retargeted to `main` with #502's commits behind it, so no rebase was needed; five harness-only commits under `deploy/mobile/`. Draft flag read as a formality. `v0.2.0` stays a draft.

PR #514, opened 2026-09-17
tests: the two remaining scripts that assumed a done wake — archive_children_e2e waits on its worker, feedback_bundle_e2e keeps its worker behind root's first turn

unarbos:
## What

- `archive_children_e2e`: root's spawn is `wait: true`, so the worker's report is in the spawn result and one turn ends whichever finishes first. The read of the archived record is the test's own question a turn later; the `archive_children = false` case asserts the folder is still in place after root's single turn.
- `feedback_bundle_e2e`: the worker's bash gains `sleep 2`. This test asserts a done wake *inside* the exchange — that is the shape under test — so the worker must report after root's first turn ends; the sleep keeps it there on a loaded runner, and the comment says so.

## Why

The fold race reddened `helper_kinds_e2e`, `multitask_e2e` and `archived_children_e2e` today: a worker's report landing while root's turn runs folds into that turn as a `say` and opens no second turn, so a script that scripted the "done turn" waits for it in vain, and the next user question reads the wrong line. I read every kernel test that spawns for a turn-count wait or a done-wake assertion; these two were the last with the shape. The rest wait on facts, or script a worker that sleeps.

## How to check

- [x] `cargo test -p arbos-kernel --test archive_children_e2e` (4 pass, three runs); `--test feedback_bundle_e2e` (2 pass)
- [ ] `cargo test --workspace` on CI

## Notes for the reviewer

Test-only. Independent of #513; the `archived_children_e2e` fix there is the same shape.

<div><a href="https://cursor.com/agents/bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #513, opened 2026-09-17
remote/acp: the last four swallowed writes on the audit's list are said

unarbos:
## What

Log lines, each saying what the reader will believe if the write did not land:

- `remote::forget` — `remotes.json` not saved: `remote_record_kept`, the record of a child that no longer exists stays until the next start drops it (`restore` already skips a record whose agent folder is gone, so nothing restarts over there).
- the stand-in's fixed address (`remote.rs:871`): `remote_address_unsaved`, after a restart the child is reachable only by its old address.
- `remote::mirror` — the mirrored lines not appended: `remote_mirror_unwritten`; the mirror mark not saved: `remote_mirror_mark_unsaved`, a restart mirrors the same remote lines a second time (a repeat, not a loss).
- the ACP session id (`acp_worker.rs:381`): `acp_session_unsaved`, the next turn opens a fresh session.

Plus one test fix the first CI run surfaced (unrelated to the change): `archived_children_e2e` waited for root's *second* turn — the done wake — but a worker that finishes before root's next step folds its report into the running turn and opens no second one, so on a loaded runner the wait ran out. Root's spawn is now `wait: true`; the report is in the spawn result and the script reads the same in either order. Same family as #463/#474/#483.

## Why

`internal/unchecked-writes-and-orderings-audit.md`: the remaining "listed, features" rows after #509. `blocked.rs`'s write side already says its failure (`blocked models: …`), so it needed nothing. With this, every row on the misleading list that was the kernel's has been either fixed or made to speak.

## How to check

- [x] `cargo test -p arbos-kernel --lib remote::` (7); `remote_track_e2e` (2), `remote_kernel_e2e` (3)
- [x] `cargo test -p arbos-kernel --test archived_children_e2e` (3 runs green)
- [ ] `cargo test --workspace` on CI

## Notes for the reviewer

No behaviour change when the writes land. The two mirror cases are the only ones a person could notice (doubled remote lines after a restart); the log now names the cause beside the symptom.

[...]

PR #505, opened 2026-09-17
gate: R24–R27 — clipped is not-reachable, the Project page rows drive the drawer, the settings walk scrolls, a Stop the kernel holds stops the cascade

unarbos:
Rig corrections from cycles 36–37, all `desktop/qa/parity/ui_pass.py`.

**R24 — clipped is not-reachable.** The driver refuses to click what is not on screen (R13); `check()` knew `no element` / `not reachable` / `occluded` as the unreachable class but not `not on screen … clipped away`, so two settings controls below the fold read as fails.

**R25 — the Project page rows had gone quiet.** Since the side-panel rewrite the page is the drawer's Project tab, closed by default; every `project-page-*` row read `not-reachable` and said nothing louder. The rows now open the drawer and drive its ways out — Escape closes it, the close mark closes it, the Project tab is first and active; Expand is recorded with the widths (280 → 280 at 1100 wide), not judged.

**Scroll before touching.** `reveal(el, body)` scrolls the settings body (or the drawer) until a control is on screen; `reveal-0/1/2` became rows. The panel phases reveal or skip a row scrolled out of the drawer instead of crashing on it (R26: phases R and W died on `panel-agent-27 … clipped away` after a run that opened many chats). The `reveal-*` controls themselves are now skipped: they open the system file manager, which then sits over the display (five Thunar windows after one run).

**R27 — a Stop the kernel holds stops the cascade.** One kernel event — a root waiting on a spawn, Stop held until the child returns (`features-inbox/2026-09-17-stop-waits-for-a-blocking-spawn.md`) — turned into 17 fails in one run: every `recover` failed the same way and every later phase ran in a chat that could not be stopped. `recover()` now names who is busy in the kernel's own words and, on the second failure, marks the run as one the kernel is holding; the main loop records each following phase `not-reachable` with that reason (`fault=False`) instead of failing rows that measure nothing.

## Gates (main `dbad816`, build 1657, kernel `arbos-kernel 0.2.0 7e19f9e90947 protocol 1`)

- `cycle-37` and `cycle-37b`: the kernel held Stop in both (70 · 17 and cascading); an A/B of phase T on builds 1616 and 1657 back to back passed both (23 · 0, 24 · 0) — the event is the kernel's, not the build's.
[...]

PR #516, opened 2026-09-17
audit_kernel_2_e2e: the nudge test's first helper sleeps so its done turn is real, and the second is waited on

unarbos:
## What

`a_coordinator_that_spawns_and_leaves_the_page_alone_is_nudged`:

- The first helper runs `bash sleep 4` before its one word, so its report lands after root's spawn turn has ended and opens the done turn the test reads (the spawn turn nudges; the done turn, also leaving the page alone, does not repeat it).
- The second spawn is `wait: true`: one turn, no done wake; the round asserts what it is about — the user's next words re-arm the nudge, and only once — and the turn count is pinned at three.

## Why

The red on #505, and the last done-wake-shaped script in the kernel suite after #463, #474, #483, #513 and #514. Its helpers answered in a millisecond; on a loaded runner the report folded into root's running turn, no done turn came, and the scripted `noted` line shifted onto the next user's turn, so the second round saw one nudge instead of two.

The first round genuinely needs a done turn (that is the "once per idle period" half of the rule), so it keeps the wake and makes the worker outlive root's turn by a margin, as `feedback_bundle_e2e` now does; the second round does not, so it takes the deterministic shape.

## How to check

- [x] `cargo test -p arbos-kernel --test audit_kernel_2_e2e` (3 pass); the nudge test 3 sequential runs and 8 concurrent under `yes` load, all green

## Notes for the reviewer

Test-only. All helper and root lines are pinned by agent, so #481's rule applies and no line can be taken by the wrong agent.


unarbos:
Merge steward: merged — `main` is now `d2015bff`; `rust` fast-forwarded. Test-only, taken as a flake fix (the red on #505); draft flag read as a formality. `audit_kernel_2_e2e` 3 pass on the merged tree. `v0.2.0` stays a draft.

PR #517, opened 2026-09-17
dev channel: a run does not wait on the commit that started it

unarbos:
The channel sat still from 21:37 to 22:19 on 2026-09-17. Four dev-channel runs in a row decided to wait, and each waited on **its own triggering commit**:

```
##[notice]df2df2c is newer and still running. Its run decides, and comes back here whichever way it goes.
```

`df2df2c` was that run's own `workflow_run.head_sha`. There was no future event to come back.

## Why

The plan asked the API what CI had concluded for every commit on its walk, including the one whose CI completion started the run. That is the one commit it must not ask about.

A commit gets **two** CI runs, about forty seconds apart:

```
created 21:47:10Z  updated 21:56:23Z  completed/success
created 21:47:09Z  updated 21:57:05Z  completed/success
```

The first to finish fires the event. The API a moment later still shows the sibling going, so `ci_state` answers `pending`, and the plan waits for something that has already happened.

## The change

The event already carries that run's conclusion, so use it rather than re-asking. `api_state` is the API question; `ci_state` is the same question except that the triggering commit can never be answered "pending".

A sibling that went green still counts — a failed trigger asks the API in case one did — but the answer for that commit is never *wait*, because there is nothing left to wait for.

## Tested

Three cases added to `.github/dev-channel-plan-test.sh`, each checked by disabling the guard and confirming it fails:

```
  ok    the run does not wait on the commit that started it
  ok    a failed trigger still sees a sibling run that went green
  ok    a failed trigger with nothing green on it falls back
```

Without the guard the first reports `expected 4, decided none` — it waits, exactly as production did.

## Two faults in the harness, found by writing them

Both had been making the suite agree with whatever it was given, which is worth naming because the suite is the only thing standing behind this decision:

- The new default took the event's conclusion from a variable that is **set-but-empty** in the ordinary case, and `${VAR-default}` substitutes only for an *unset* name. Every case read `success`, so the five that assert a fallback silently stopped asserting anything. `:-` fixes it.
- The "commit with no CI run" case had the triggering commit be the one with no run — which cannot happen, since a commit with no run fires no event. It is now a commit the walk passes over, which is the shape that actually occurs.

Eleven cases pass.

## Not changed

`v0.2.0` remains a draft.


unarbos:
Merge steward: merged — `main` is now `077ed50b`; `rust` fast-forwarded. Read on the publish path: the triggering commit's answer comes from the event's own conclusion (`github.event.workflow_run.*`), no new permission, no secret. Hermetic test 11 of 11 on the merged tree, `all good`. Draft flag read as a formality. One observation for the record: the "sibling run forty seconds apart" is the same sha's CI run on `rust`, which the steward fast-forwards right after `main` — so the double run is structural, and this guard is the right fix rather than a timing one. `v0.2.0` stays a draft.

PR #521, opened 2026-09-17
turn: the wait for the checkpoint tree is bounded — past 20 s the turn goes on without a file checkpoint, says so once, and a tree finishing later is dropped rather than kept

unarbos:
## What

The turn's first writing tool waits for the checkpoint of the working tree (#405, qal-j17). That wait now has an end: `tree_wait`, 20 s (`ARBOS_TREE_WAIT_MS` for tests). Past it:

- the hold is released and the tool runs;
- the tree that finishes later is **dropped**, not filled in — the record's `work_error` says *the working tree took too long to snapshot and the turn went on without it*, so `rewind --files` to that turn refuses with the reason (the transcript rewind still works);
- root's transcript says once per place what happened, the usual cause (a large repository; untracked build output or `node_modules`), and the fix (`.gitignore` them).

## Why

On a large repository `add -A` takes seconds, which qal-j17's fix accepted. On a machine we did not choose — a home folder that is a dotfiles repository, a monorepo with build output untracked — it runs for minutes, and every turn's first command would have sat behind "Saving a checkpoint of the working tree" for as long as it took. A person's command should not wait on a folder a checkpoint cannot keep up with.

Why drop the late tree rather than keep it: it was taken beside the turn's own writes, so it may already hold them, and a correct-looking restore of the wrong state is the worst outcome in this family — worse than a refusal, worse than nothing (qal-j17's own words). The record's honesty is what makes the bound safe.

## How to check

- [x] `cargo test -p arbos-kernel --test tree_wait_bound_e2e`: a 4 s tree against a 1.5 s bound — the write starts under 3.5 s, `f1.txt` exists, the notice is on the transcript once with the `.gitignore` advice, the late tree lands as `work_error` with no `work`, and a second turn does not repeat the notice.
- [x] `rewind_no_identity_e2e` (the 2 s tree under the default bound still waits and the rewind never restores the turn's own file), `standing_pass_e2e`, `confirmed_reads_e2e`; `cargo test -p arbos-engine --lib tools::git` (17).

## Notes for the reviewer

- `snapshot_turn_tree` keeps its signature; `snapshot_turn_tree_unless` takes the abandoned flag. `TREE_TOO_SLOW` is the record's wording.
- The bound task and the snapshot task share one `watch::Sender`; whichever finishes first releases the hold, and the flag decides whether the tree is kept.
- 20 s is a judgement: long enough that a normal large repository never hits it, short enough that a person notices the wait rather than the absence of a checkpoint.


unarbos:
Merge steward: merged — `main` is now `b4971f92`; `rust` fast-forwarded. Read with the qal-j17 care. The property that matters holds under the race: a tree is kept only when `work_commit` finished before the guard read `abandoned == false`, and the hold is released only after that guard passed (tree is pre-turn) or after `abandoned` was set (tree dropped, `work_error` recorded, file rewind refuses). The one narrow window — the bound firing between the guard and the record write — drops a good tree, which is the safe direction. On the merged tree: `tree_wait_bound_e2e` 3, `rewind_no_identity_e2e` 3, `standing_pass_e2e` 1, `tools::git` 17 pass. Draft flag read as a formality. Mac worker: build from `b4971f92`. `v0.2.0` stays a draft.

PR #524, opened 2026-09-17
loop: six journey checks that could not pass, or failed in silence

unarbos:
Found by running the acceptance journey to prove #515, which merged while that run was in flight. Every one is the same shape as the fault that PR fixed: a check that could not pass, a failure that said nothing, or a tap that landed somewhere other than where it reported. Together they were three of the journey's standing failures.

## J1 scored FAIL on every run

`wait_hist` reads the transcript through `hist`, which refuses to read without an anchor. The anchor is `AFTER`, and `AFTER` comes from **the very line being waited for**, set on the next statement. So the wait had nothing to read against and timed out every time. J1 then passed on the real check a few lines later, which is why nobody chased it. It waits on the line's own seq now:

```
J1 PASS the setup line reached the kernel after 0s, at seq 2122
```

## The recording produced no file, and nothing said so

A recorder left behind by an interrupted run holds the device, and every later `recordVideo` is refused with `Host recording is already in progress`. The journey sent that output to `/dev/null`, so the only hint was ffmpeg failing on a missing input several lines later. The loop's owed recording was partly this. Stale recorders are cleared first, the refusal is logged, a missing file is said out loud. Worth knowing: killing the process is not enough — a wedged recorder needs `simctl shutdown` and `boot`.

## A tap could land on something not on the screen

`describe-all` returns the whole scroll view, including messages far above the top. `ui tap "Up"` found the letters `up` inside the word **`setup`**, in a message at **y=-573**, and tapped that — reporting success. Matching is now confined to the visible frame and the substring fallback needs a whole word. The step that used to lie now says `nothing matching 'Up' on screen`, which is how the next fault surfaced within a minute.

## P1 could never have sent

The one button right of the composer is three buttons in turn: **Microphone**, **Stop** while it listens, **Up** once there are words. The step tapped `Microphone` twice, and by the second tap there was no Microphone there. Dictation itself was never broken — the box held the words every time. Proven end to end now:

```
2167 user       Please summaries what the workers did today in two sentences.
2168 assistant  Today, workers completed creative writing tasks …
```

## P2 and P3 were one fault

P2's photo tap was at y=230 **points**, which is the "Private Access to Photos" banner, not the grid. Nothing was selected, so the picker's tick stayed disabled, so the picker never closed — and it swallowed the photo line **and the entire call step**. The last run's `P3-call.png` and `P3-call-answered.png` are stills of the photo grid. It went unnoticed because P2 scores on the kernel's reply, and the model answers a question about a photo it cannot see, plausibly.

The way into the picker is by name (`Add` → `Photo Library`). The taps inside it must be points — it is another process and `describe-all` cannot see in — so they go through `sim-lib.sh`'s `tap_shot`, which converts from screenshot pixels. **The way out is now checked**, and a picker that had to be dismissed by hand scores P2 as *no photo attached* rather than quietly asking about nothing.

P3 opens the call from the chat menu by name, and its question lands in this project's transcript, which is the thing P3 exists to check:

```
2175 user       Hello Arbus. What are we working on right now? Give me one sentence.
2176 assistant  We just finished the J224840 challenge — …
```

## `record.jsonl` beside `record.json`

`journey-record.py` writes `record.json` indented for reading and prints the one-line form for the ledger. Appending the indented file to `internal/mobile-journey-history.jsonl` breaks it, which is what I did to run 32's entry. The one-line form is written to disk too now.

## One of these was mine

Fixing the picker I read the tick's position off the rendered still as `426,157` and tapped that. The screenshot is 472 wide and the device is 393 points, so it landed nowhere, and the original `355,131` had been right. `sim-lib.sh` has existed for this since cycle 46; the journey simply never sourced it. It does now. A rule only holds where it is reachable from.

## The run behind this

Journey run 32 on `pod`, kernel `arbos-kernel 0.2.0 efcab58f29e1`: **7 typed lines of 7 reached the kernel**, against 2 of 8 in run 31. Every J step passes or carries a standing `unverified`. With P1, P2 and P3 fixed here, the journey has no step left that fails for a reason in the harness.

Evidence in the project store: `media/mobile/cycle-55/` (19 stills, both recordings) and `media/mobile/cycle-56/` (the photo attached, the call opened and answered).


unarbos:
Merge steward: merged — `main` is now `77f7c964` (with #469 and #433); `rust` fast-forwarded. Harness-only under `deploy/mobile/`; scripts parse. Draft flag read as a formality. `v0.2.0` stays a draft.

unarbos:
Merge steward: heads-up — `9054387e` ("the journey's photo and call steps, by name and with the picker checked") was pushed to this branch at 23:25, after `a90487ab` was merged at 23:25 as `db061b3d`. GitHub still shows this PR open because the head moved, but a merged PR whose branch keeps moving is the class the exit check flags (it is how #418's probe was lost). It is harness-only and small; I will take it on green as a normal merge of this PR, but in future please open a fresh PR for work after a merge rather than pushing to the merged branch.

unarbos:
Merge steward: the follow-up `9054387e` is merged too — `main` is now `3aca326a`; `rust` fast-forwarded. Harness-only; scripts parse. Nothing left on this branch.

PR #529, opened 2026-09-17
ios: a search that matches nothing says so

unarbos:
Found re-checking the list composer, a coverage row last exercised at cycle 11.

Search the project list for something that matches nothing and the screen goes blank under the search box — with an empty **"Read ⌄"** header sitting over it. That header reads as a section somebody collapsed, so the honest answer ("nothing matched") looks like a list that is still loading, or one you have accidentally folded shut.

The cause: the empty state only fired when the **roster** was empty.

```swift
if projects.entries.isEmpty { emptyState }
```

A filter emptying the list is a different thing and had no branch at all.

Now it says which:

```
No project matches “zzzz”.
```

and with the live-only filter on and nothing live:

```
No project is live. Turn the filter off to see the rest.
```

The empty `Read` header also stands down when a filter empties the list, so the message is not competing with a heading that implies there is something under it.

Before and after: `media/mobile/cycle-57/04-no-match-before-blank-screen.png` and `05-no-match-after-it-says-so.png`.

## The rest of the row, measured

The three standing claims for this row all hold on `77f7c964`, checked on screen rather than assumed:

| claim | measured |
|---|---|
| the placeholder names the project a line goes to | 7 rows on screen, `Message phone…` |
| with a filter, it names one that is on screen (#433) | filter `sub` → only `subnet120` → `Message subnet120…`; filter `const` → `Message const…` |
| no match offers nothing rather than an invisible project | `Plan, ask, build…`, and send stays disabled |
| the composer rides above the keyboard | composer top moves 774 → 480 points when the keyboard comes up; the keyboard starts at 683, so a **182-point** gap |

#433's rule is working: the composer follows the visible rows, not the remembered target. The keyboard claim was last checked at cycle 11 and is still true; it is now measured from the accessibility tree rather than eyeballed, so the next check is a number rather than a judgement.

Stills: `media/mobile/cycle-57/`.


unarbos:
Merge steward: merged — `main` is now `2eae41c7`; `rust` fast-forwarded. One `ios/` file, iPhone job green. Watching the TestFlight run for the build number. Draft flag read as a formality.

unarbos:
Merge steward: TestFlight run for `2eae41c7` printed `uploaded 0.2.0 (1716)` — build 1716 carries this.

PR #532, opened 2026-09-18
leash: a zombie kernel is believed only when seen twice, 200 ms apart — a process de-threading for an execve shows as Z for an instant

unarbos:
## What

The leash's `alive()` re-reads the kernel's process state 200 ms after a `Z` and treats the kernel as gone only if it is still `Z` (and `kill -0` still succeeds). A single `Z` is no longer enough.

## Why

I ran the whole kernel suite under six CPU spinners to find timing-shaped tests before CI does. One product fault came out rather than a test fault: `update_gate_e2e`'s *a job survives an exec into the kernel* failed 7 of 12 under load, with the job's `killed` marker reading **"the kernel exited and the job was ended with it"**. The kernel had not exited. A multi-threaded process that calls `execve` from a thread other than its leader is de-threaded by the OS — the exec'ing thread takes the leader's pid and the old leader is reaped through a zombie state — and a slow machine holds that state open long enough for the leash's one look at `/proc/<ppid>/stat` to read `Z`. The leash then killed every job, at the very moment the update test exists to protect.

A kernel that has really died and not been reaped stays `Z` for ever, so the case the check was built for (SWE-bench cycle 14: 4–10 live test processes per rollout after the kernel exited) is caught exactly as before, 200 ms later.

## How to check

- [x] `update_gate_e2e a_job_survives…` under six `yes` spinners: **12 of 12** (before: 5 of 12).
- [x] `job_leash_e2e` (12, including the zombie-kernel case), `update_gate_e2e` (2), `sigchld_blocked_e2e` (3), `cargo test -p arbos-engine --lib jobs` (8).

## Notes for the reviewer

- In production the kernel execs from its main thread (`rt.block_on(serve::run)`), where the leader-zombie window may not open; the test's helper execs from a libtest thread, which does. Either way the leash should not act on one glance at a state that a legitimate transition passes through.
- The whole suite under load otherwise passed: 130 test binaries, this the only red.


unarbos:
Merge steward: merged — `main` is now `8a407ca0`; `rust` fast-forwarded. Read on the kill path: `alive()` is strictly more conservative about declaring the kernel dead (a `Z` must be seen twice, 200 ms apart, with `kill -0` still succeeding), and a real unreaped zombie stays `Z`, so the case the check was built for is caught 200 ms later than before. On the merged tree: `job_leash_e2e` 12, `update_gate_e2e` 2, `sigchld_blocked_e2e` 3 pass. Draft flag read as a formality. Mac worker: build from `8a407ca0`. `v0.2.0` stays a draft.

PR #533, opened 2026-09-18
ios: the banner calls the project what the rest of the app calls it

unarbos:
Found re-exercising the notifications row, last checked at cycle 36.

A reply arriving while the phone is elsewhere raises a banner that reads:

```
arboslife/phone · root replied
OK
```

Every other surface calls that project **`phone`** — the chat header, the list row, the composer placeholder. The banner is the one place that shows the raw `machine/project` target, and it is the place with the least room and the least context.

## Why

`ChatStore.title` went straight to the target's label:

```swift
let target = settings.kernelTarget.label   // "arboslife/phone"
```

The chat view had always preferred the project's own name, then its folder, and only then the target:

```swift
identity.label ?? entry?.folder ?? target.label
```

The header therefore read `phone` while the banner drawn from the same store read `arboslife/phone`. The store follows the view's order now, and `KernelTarget` gains the `folder` accessor it needed to.

Before and after, both captured on the same simulator minutes apart: `media/mobile/cycle-58/01-banner-before-the-raw-target.png` and `02-banner-after-the-project-name.png`.

## The row itself, re-exercised

The notification path works end to end and is now a committed scenario, `deploy/mobile/scenarios/notifications.sh`: the permission ask, the banner, the badge, the tap that lands in the right project, and the away card on return. The recording is `media/mobile/cycle-58/recording-notification-away-and-back.mp4`.

## Two things the scenario had to learn, both of which faked a failure first

**The banner is SpringBoard's, not the app's.** `idb ui describe-all` cannot see it, so a detector that looked in the accessibility tree reported "no banner in 120s" through a run where one was plainly on the screen. The scenario reads the app's own console (`notify … reply` / `banner …: posted`) and photographs the Home Screen.

**A slow reply never raises one at all.** The banner is posted from the live socket, and iOS keeps a backgrounded app alive for roughly half a minute. The first run asked the kernel to count slowly to twenty, went away, and got a badge but no banner — because the reply arrived after the app was suspended, and push is still off until Jacob's APNs key lands. The scenario asks for a one-word reply, and the reason is written down beside it.

Neither is a fault in the app, but both would have been filed as one.


unarbos:
Merge steward: merged — `main` is now `74b49b4c`; `rust` fast-forwarded. Two `ios/` files plus a rig scenario; iPhone job green. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #535, opened 2026-09-18
ios: a heading is a heading and a bullet is a bullet

unarbos:
Found pairing the phone's chat against the Cursor reference stills.

The kernel writes markdown. The phone parsed only the **inline** kind, so everything that starts a line arrived with its marker showing:

```
## Shapes
1. Step one
2. Step two
3. Step three

- Note one
- Note two
```

Bold and `inline code` were fine. Headings came through as `## Shapes`. Bulleted lists came through as `- Note one`. Numbered lists were the only kind that looked right, and only because `1.` reads as a number whether it is parsed or not — nothing was rendering them either.

## Why it was written that way, and why that part stays

```swift
AttributedString(markdown: text, options: .init(interpretedSyntax: .inlineOnlyPreservingWhitespace))
```

`inlineOnlyPreservingWhitespace` is the right parser for this chat: the full one reflows the text and throws away the model's own line breaks, which is worse than a visible hyphen. So rather than change parsers, the markers that start a line are turned into the typography they stand for **before** the inline parse — a heading becomes bold, `- ` becomes `•  `.

Lines inside a fence are left exactly as written. A `-` in there is the code's, not the prose's.

Before and after, same prompt, same kernel:

| | |
|---|---|
| `media/mobile/cycle-59/01-markdown-before-raw-markers.png` | `## Shapes`, `- Note one` |
| `media/mobile/cycle-59/02-markdown-after-bullets.png` | `•  Note one` |
| `media/mobile/cycle-59/03-markdown-after-heading.png` | **Shapes** |

The kernel's text was identical across both runs (seq 2210 and 2214), so the difference is the rendering and nothing else.

## The rest of the pairing, for the record

**Links are already right.** A markdown link (`[#231](…)`) and a bare URL (`https://arbos.dev`) both draw in the accent colour and are tappable, as Cursor's `#231` does — `media/mobile/cycle-59/06-links-and-a-wrapped-list-item.png`. Checked rather than assumed; neither is something this app maintains.

**A long list item wraps to the margin where Cursor hangs it**, so after the first line the list's shape is gone. Left as it is, deliberately: a hanging indent costs about twenty points of width on every line of every list item, and this is a phone — Arbos buys text per line where Cursor buys structure, and the item boundary still reads because `2.` starts a line. It is also a renderer restructure rather than a setting, since the agent bubble is one `Text` with the streaming caret on its baseline, which is not worth doing on a judgement call.

**The lists agree on shape** — same top bar, collapsible `Working` / `Read` sections, row of glyph, name and a second line, composer pill with `+`, placeholder and mic. Arbos's placeholder names the project (`Message phone…`) where Cursor's is generic; the better of the two, and deliberate.

**One difference I could not close.** Cursor puts the time since each agent last did something on the right of every row — `1m`, `4m`, `3m`. Arbos shows nothing, so seven rows reading `Idle` are indistinguishable and the list offers no sense of what moved recently. The data is not available to the app: the hub's `/list` gives each project `access`, `identity`, `live`, `name`, `place`, `share`, `store` and no timestamp at all, and the only time on the payload is `since` on the machine. Filed for the hub's owner rather than guessed at.


unarbos:
Merge steward: merged — `main` is now `3ef5f436`; `rust` fast-forwarded. One `ios/` file, iPhone job green; draft flag read as a formality. Watching the TestFlight run for the build number.

PR #537, opened 2026-09-18
ios: Settings stands on the app's own ground

unarbos:
The second finding from the style pair against the Cursor reference stills. #535 carried the first.

Sampled rather than eyeballed, because the difference is small enough to argue about:

| screen | dominant colours |
|---|---|
| projects list | **92.3% `#161514`**, 2.8% `#212121` |
| settings, before | 48.4% `#161514`, **45.4% `#2c2c2e`** |
| settings, after | 48.4% `#161514`, **45.4% `#212121`** |

`#2c2c2e` is iOS's grouped-cell grey in dark mode. It appears nowhere in this app's palette. Nearly half of one screen was standing on a colour the app never chose, while every other screen is `#161514`.

## Why it slipped through

The view already did the obvious thing:

```swift
.scrollContentBackground(.hidden)
.background(ArbosTheme.bg)
```

That hides the page behind a `Form` — but not the fill behind each **row**, which keeps the system's grouped-cell colour. Applying `.listRowBackground` to the `Form` does not reach them either; it has to go on each section.

The rows now take `ArbosTheme.card`, the raised colour the composer and the Agents pill already use, with the page still `ArbosTheme.bg`.

Before and after: `media/mobile/cycle-59/07-settings-before-ios-grey.png` and `08-settings-after-the-app-palette.png`. The eye barely registers the change, which is why it survived fifty-nine cycles and why a pixel sample found it.

## Related, checked, and left alone

**Dark-only is deliberate on both sides.** The reference stills are light, which raised the question of what the phone does there. It forces `.preferredColorScheme(.dark)` with a fixed hex palette; `desktop/src/view/palette.rs` uses `cursor_dark` with `t.bg = grey(0x161514)` — the same literal value as `ArbosTheme.bg` — and says outright that Cursor's light theme was never sampled. Same decision, same number, nothing to change. A phone that quietly followed the system into light mode would have had no palette to do it with.


unarbos:
Merge steward: merged — `main` is now `ccca6841`; `rust` fast-forwarded. One `ios/` file, iPhone job green; draft flag read as a formality. Watching the TestFlight run for the build number.

PR #539, opened 2026-09-18
loop: a scenario for the four rows that had gone longest unchecked

unarbos:
Cold start, away-and-back, long history and attachments had all been sitting at cycle 37 — twenty-three cycles. This measures them, and the numbers are the point: each row had a figure from cycle 37 to beat or match, and now has one for next time.

## What the run says

| | cycle 37 | now | |
|---|---|---|---|
| cold start to a list with rows | 3.1 s | **3.0 s** | seven rows |
| tap to a chat with a composer | 1.1 s | **0.9 s** | on **2220** transcript lines, where cycle 37's was 1265 |
| away 8 s and back | "chat as it was" | **26 text rows before, 26 after** | |
| `Show 200 earlier lines` | "pages 200 back" | **3.6 s**, older content arrives, position holds | |
| photo chip | "chip in the field, send arrow replaces the mic" | chip with its `×` present, right-hand button is `Up` | |

Nothing has slipped, and the chat opens *faster* on nearly twice the history. Stills in the project store at `media/mobile/cycle-60/`.

## The scenario, and the trap it walked into first

The first version flung six times to reach the top of the window and reported:

```
pager: not found
```

The pager is real. It sits at the top of a two-hundred-line window of long replies, which thirty flings reach easily and six do not — and six had no way to know they had fallen short. Filed as an app regression that would have cost a cycle chasing a control that works.

It now flings until the pager appears **or the transcript stops moving**, and says which happened:

```
pager after 10 flings: Show 200 earlier lines
paged in 3.6s; the tail of the window is now harness probe 220556-2
```

A rig that cannot tell "not there" from "I did not look far enough" should not be allowed to say either.

## Why measure the tree rather than the screenshot

Every number here is a poll of the accessibility tree. The tree is what says the rows exist; a screenshot says only that pixels changed, and this loop has been caught by that difference more than once.


unarbos:
Merge steward: merged — `main` is now `de2eff6e`; `rust` fast-forwarded. Rig-only (one new scenario script; parses). Draft flag read as a formality. No app code, so TestFlight stays 1735.

PR #541, opened 2026-09-18
prompt: "no change needed" takes evidence; producer rule covers the copy and the guard

unarbos:
## What

Inbox item `2026-09-18-swebench-no-change-rule-and-producer-scope.md` from the SWE-bench loop (cycle 23 evidence, cycle 22 runs).

**A sixth rule, beside the done criterion.** django-13513, four of four rollouts in both arms: the agent found the *issue's suggested patch* already in the tree, ran a test covering that suggestion, and said no change was needed. The hidden test wanted the *symptom* fixed. Two earlier cases declined after reading upstream history. New `CONTRACT` line:

> Concluding that the request needs no change needs the same evidence as a fix: run the request's own example against the unmodified tree (`bash repro:true`, so the run is on record) and show it already behaves as asked, with the command and its output in your reply — one line, `"no change: <command> already <does what the request asks>"`. "The fix the issue suggests is already present" is not that evidence …

Visible marks: a recorded `repro:true` run on the unmodified tree, and a `no change:` line in the reply. Both are greppable in a transcript.

**Producer rule's scope widened.** xarray-6938 and sympy-17318: the agent's own root-cause line named the producer ("`to_index_variable()` returns `self`"; "`_sqrt_match`'s condition") and the diff went to the consumer. Neither matched the rule's examples (attribute/case a sibling has). The rule now also names *a method that returns the same object where a copy is needed* and *a wrong value made in one place and checked where the error surfaces*, and says the root-cause line in the reply names where the fix goes.

## A bug the longer prompt surfaced (second commit)

`budget_once_e2e` went red on the first push: the over-budget notice was said twice in one turn (`~4k tokens` then `~5k tokens`). `compact::said_this_turn` compared notices whole, and the line carries the working set's size, which grows between steps; the longer standing prompt moved the count across a thousand mid-turn. The guard now keys on the words before the first measurement (` (~`). Unit test in `compact.rs`; `budget_once_e2e` green locally.

## Tests

- `prompt.rs`: new `saying_no_change_is_needed_takes_a_recorded_run_and_a_line_in_the_reply` pins the marks and that the rule sits directly under the done criterion; the existing twin/stage/producer test pins the widened wording.
- `compact.rs`: `the_over_budget_line_is_one_notice_whatever_the_count_says`.
- `cargo test -p arbos-engine --lib`: all pass. Clippy clean.

## Not changed

The django-16877 note (sibling not opened) is the existing producer rule not being followed; no wording change asked, none made.


unarbos:
Merge steward: merged — `main` is now `2492c601`; `rust` fast-forwarded. Prompt wording plus a dedup key for the over-budget notice that ignores the measurement suffix. Engine lib and `budget_once_e2e` pass on the merged tree. Draft flag read as a formality. `v0.2.0` stays a draft.

PR #538, opened 2026-09-18
hub: per-project last_activity_ms on the roster

unarbos:
## What

Inbox item `2026-09-18-hub-last-activity-per-project.md` (iPhone loop, cycle 59): `GET /list` carried no time per project, so the phone's list showed seven rows all reading `Idle` in alphabetical order. Cursor's list shows `1m` / `4m` beside each row and sorts by it.

## How

- **Frame** (`arbos-core::hub`): `HubFrame::Activity { project, at_ms }`, kernel → hub.
- **Kernel**: `Hooks::tell_activity()` sends it from `turn_started` and `turn_ended` (any agent of the project). At registration the link sends one more from the top-level transcript file's last write, so the roster has a value before the first turn.
- **Hub**: keeps the newest per project (`MachineEntry::activity`); an older or repeated time changes nothing. Each `ProjectInfo` carries `last_activity_ms` (Unix millis; absent until a kernel has spoken for the project). The roster is broadcast when the value moves forward, so an attached phone updates live.
- Design doc: one line under the roster's `ProjectInfo` fields.

Shape for the phone: `projects[].last_activity_ms`, subtract from now. Sort descending for most-recent-first.

## Tests

- `arbos-hub` unit test `the_roster_says_when_each_project_last_moved`: newer wins, older/repeated/zero do not move it, the row serialises the field, a project nobody spoke for carries none.
- `cargo clippy --tests` clean on the three crates (pre-existing engine warnings untouched).

<div><a href="https://cursor.com/agents/bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #480, opened 2026-09-17
ios: a project he has opened does not vanish when its machine goes off

unarbos:
Cycle 51 opened this as a draft with half of it working. Cycle 52 finished it, and the answer was not what I expected.

## Why

Stop a project's kernel and its row simply disappeared from the list. Correct by the old rule — the roster lists registered machines — and from Jacob's side it reads as his work having been lost rather than being asleep.

It also hid the only place the app explains itself. The refusal wording from [#420](https://github.com/unarbos/arbos/pull/420) and [#465](https://github.com/unarbos/arbos/pull/465) needs a row to tap, and the row went before he could tap it.

## The change

A project he has opened at least once keeps its row when the roster stops listing it, drawn Off and saying **which** thing is missing. The roster answers that too, so the line is specific rather than a shrug:

```
c52-place
awsmac is off · fieldwork
```

`awsmac is off` when the machine is not registered at all; `<project> isn't running on <machine>` when the machine is there and the kernel is not. Only projects he has opened are kept, capped at forty. M-88's behaviour for a hub that does not answer is untouched.

## What the draft was waiting on

The row survived but was not named, and survival alone did not say which code path kept it — M-88 keeps rows too. Rather than guess a third time, the pass now says what it decided, and the answer came out at once:

```
roster: kept-row pass — listed 1, opened before 1, machines []
roster: keeping hub:awsmac/fieldwork — awsmac is off
```

**Nothing was wrong with the branch.** Cycle 51's runs simply had nothing remembered yet — `opened before 0` — because the project had not been opened in that install before its kernel was stopped. The fix worked the whole time; the test did not exercise it.

One debug line stays, the one that says which path kept a row and why. Reading that off the screen is exactly what sent the last cycle guessing.

## Verified on the final commit

Built from `06e9f646`, run end to end: row `c52-place, awsmac is off · fieldwork`, trace `roster: keeping hub:awsmac/fieldwork — awsmac is off`.

Evidence in the project store under `media/mobile/cycle-51/` and `media/mobile/cycle-52/`.

[...]

PR #543, opened 2026-09-18
ios: three things the list and chat were not saying

unarbos:
Three changes from cycle 60, which worked through the coverage rows that had gone longest unchecked.

## 1. The list lost six projects and said nothing

Save a hub token that cannot work and the list falls from seven rows to one — the pod's own kernel, which needs no hub. It gave no account of why.

Everything needed was already there. `HubError` words it plainly (`Hub token refused.`) and `ProjectStore.problem` holds it, but the only thing that ever rendered it was `emptyState`, which fires when `entries.isEmpty`. The list is never empty, so that sentence was unreachable.

Now it shows whenever the hub has something to say: `media/mobile/cycle-60/14-list-after-a-bad-token.png`. Same shape as the empty-search fix in #529 — a message that existed, with only the all-empty case able to reach it.

## 2. How long since each project did anything

Cycle 59 paired the list against Cursor's and found the one thing it could not do: Cursor times every row, Arbos showed seven rows reading `Idle` in alphabetical order. The hub had no per-project timestamp to give. #538 added one, citing that finding in its own doc comment.

The row now reads `now`, `4m`, `2h`, `3d` on the right. **Absent is not zero** — when the hub has heard nothing the row says nothing rather than guessing at "just now".

**Not shown working end to end.** The hub deployed on ArbosLife still predates #538 and omits the field (`last_activity_ms= ABSENT` on every project), so the only case observable on a simulator today is the absent one: the list unchanged, no times, no guesses. That much is verified. The times themselves want a run once the hub redeploys.

## 3. A worker named two ways on one screen

Four workers spawned together showed as three `Done say sentence about mountains` and one `Done say-sentence-about-rivers`. The kernel spawned all four with spaced names (seq 2225–2228); the app keeps a map from child id to brief and falls back to the id when it has none, and the id is the goal **slugged**.

The fallback now reads the id as words.

**Observed once and diagnosed from the kernel's record; not re-observed after the fix.** The fallback fires only when a done arrives before its spawn — a race I cannot force — and these lines are live-only, never replayed from history, so a relaunch cannot show them either. Stating that rather than implying a verified before-and-after.

## The rows this came out of

| | cycle 37 | now |
|---|---|---|
| cold start to a list with rows | 3.1 s | **3.0 s** |
| tap to a chat with a composer | 1.1 s | **0.9 s**, on 2220 lines where 37 had 1265 |
| away 8 s and back | "chat as it was" | 26 text rows before, 26 after |
| `Show 200 earlier lines` | "pages 200 back" | **3.6 s**, position holds |
| photo chip | "chip in the field, send arrow for the mic" | both present |
| four workers at once | "Agents 3 → 7, sheet lists all seven" | pill **2 → 6**, sheet lists all six by name, archived children kept |

Nothing had slipped; the chat opens faster on nearly twice the history.

## Three scenarios, and what each got wrong first

- `cold-start-and-history.sh` flung a fixed six times and reported `pager: not found` on a pager thirty flings reached easily. It now flings until the pager appears **or the transcript stops moving**.
- `settings-and-a-bad-token.sh` recovered by reinstalling, which does **not** undo a saved token — it is in the Keychain, which outlives the app. It stranded the simulator once; it now restores by typing the build's own token back.
- `several-workers.sh` counted the sheet with `grep 'Done '` and reported **0** rows against six on screen: the labels read `<goal>, Done`, with nothing after the state.

Each one asserted a shape the screen never promised. All three now say which of "not there" and "I did not look properly" they mean.


unarbos:
Merge steward: merged — `main` is now `53dffd33`; `rust` fast-forwarded. Four `ios/` files plus three rig scenarios (parse); iPhone job green; the #539 commits on this branch were already on `main`. Draft flag read as a formality; the unverified items are stated as such. Watching the TestFlight run for the build number.

unarbos:
Merge steward: heads-up — `92277a70` ("leaving a call you never started no longer kills the app") was pushed to this branch at 01:31, five minutes after `7dc2d40c` merged as `53dffd33`. This PR is closed, so that commit has no PR and will not land on its own. It reads as a crash fix, so please open a fresh PR for it rather than pushing here; I take it on green. (Third time tonight on this worker's branches: #480/#485 orphans, #524, now this.)

PR #547, opened 2026-09-18
ios: a crash on leaving a call, and a workers sheet showing two of fourteen

unarbos:
Two findings from cycle 61, both on paths a person takes.

## 1. Leaving a call you never started killed the app

Open a project's call, change your mind, tap the cross. The process dies.

```
*** Terminating app due to uncaught exception 'com.apple.coreaudio.avfaudio',
    reason: 'required condition is false: NULL != engine'
```

Not the call screen — the app. It is an uncaught `NSException`, so there is no recovery and no message.

`start()` attaches the player to the engine and installs a tap on it. `stop()` removed that tap unconditionally, and a call that never started never ran `start()`, so the player has no engine and `removeTap` raises. Guarded on the player actually having one. The other teardown calls were already safe: `engine.inputNode.removeTap` sits behind a `converter != nil` check, and `player.stop()` / `engine.stop()` are documented no-ops on an engine that never ran.

Isolated to the step, not guessed at — a script that enters the call from the chat, never taps the orb, then closes:

| | app afterwards | uncaught exceptions |
|---|---|---|
| before | **GONE** | 2 |
| after | alive | 0 |

Muting was never involved; that button is correctly disabled until a call is up. I checked both steps separately.

**How it surfaced** is the part worth keeping. The scenario was not looking for a crash — it was settling a disagreement in the coverage ledger (cycle 38 said closing a call returns to the chat, cycle 56 saw the projects list), so it asked *where close landed* and printed the answer. The answer came back empty. A run that tapped the cross and moved on would have reported a clean pass.

Both old notes were right, incidentally: cycle 56 entered by launch argument, with no chat behind it. From the chat's own menu, close returns to the chat.

## 2. The workers sheet showed two of the kernel's fourteen children

Reopened after a relaunch, the sheet listed two workers. Asked directly, the kernel's `tree` frame carried twelve — including the four run an hour earlier.

`publishWorkers()` filtered on a `touched` set, which is only what *this session* watched go by. The tree's children reach `workers` and `workerOrder` through `remember()` but never `touched`, and none of them are running, so the sheet could show whatever the history replay happened to brush against and nothing else.

I first wrote this up as needing a product decision — which of a kernel's children belong on that sheet. **The desktop had already answered it.** `desktop/src/view/panel.rs` draws the live tree and folds the kernel's *archived* workers under a collapsed `N archived` row, "faint with their checks, below the live ones". It does not drop live ones. The phone's twelve were all still in the tree, so dropping ten was simply wrong.

A child the kernel has told us about is now on the sheet. Measured against the kernel's own tree: **two rows before, fourteen after**, with the four recent workers back among them — `media/mobile/cycle-61/06-workers-sheet-shows-the-tree.png`.

A refused spawn is still gone: that path removes the worker from `workers` and `workerOrder` outright, so it cannot return through this.


unarbos:
Merge steward: merged — `main` is now `e767633d`; `rust` fast-forwarded. One-line guard plus a rig scenario; iPhone job green; fresh branch against `main` — thank you. Watching the TestFlight run for the build number.

unarbos:
Merge steward: the follow-up `bb76bc31` (workers sheet shows the kernel's children) is merged too — `main` is now `9bb49c61`; `rust` fast-forwarded. The crash guard `45f0d199` had landed as `e767633d`. Nothing left on this branch. Watching TestFlight.

PR #548, opened 2026-09-18
kernel: job_stop frame — a window stops a job through the kernel's own kill

unarbos:
## What

Side-panels design, handover item 4: *Stop a job from a client, routed through the kernel's own group kill, with the reason written by a single writer.* The drawer's job row has a Stop that until now could only put words in the composer.

## How

- **Frame** (`arbos-core::wire`): `job_stop { agent, id }`, client → kernel.
- **Kernel** (`serve.rs`): finds the job under the agent; `JobsRoot::kill_saying(job, "killed: stopped by the user from the window")` ends the whole process group. A job already over is answered with its final `job` frame (`running: false`, `exit`) so the row settles; an unknown agent or id is an `error` frame naming it. Readers may not send it (the existing role gate lists what a reader may).
- **One writer for the marker** (`jobs.rs`): `kill` now delegates to `kill_saying(job, line)`, which writes the `killed` marker before the signal exactly as before. The window's stop carries its own line instead of changing the process-wide `set_kill_reason` under every other kill in flight — the "do not add a third writer" note in the design.
- **No new end path.** The job's end travels what every end travels: the `board` close with `status_line()` (now `killed: stopped by the user from the window after 3s`), the agent's kernel wake with the same words, the `surfaces` row.

## Tests

`job_stop_frame_e2e`: a background ticker; an unknown id is an error and nothing dies; Stop ends the whole group, the marker reads the one line, the `board` close carries it, the agent's transcript hears it; a second Stop on the ended job returns the final `job` frame with no exit code. Green three runs; clippy clean.

## For the desktop

Send `{"type":"job_stop","agent":"<owner>","id":"<job id>"}` from the row's Stop. The row's end arrives as it does today. The design doc's handover list (item 4) can be struck through by its owner.


unarbos:
Merge steward: merged — `main` is now `9bb49c61`; `rust` fast-forwarded. Read on the kill path: `kill` delegates to `kill_saying` with the same `kill_reason()`, so existing callers are unchanged; the new frame is in the reader-denied gate; the kill is one job's group under one agent's root. `job_stop_frame_e2e`, `job_leash_e2e` 12 and core `contracts` pass on the merged tree; desktop checks. Draft flag read as a formality. `v0.2.0` stays a draft.

PR #551, opened 2026-09-18
loop: two measuring tools that were answering the wrong question

unarbos:
Both of these are instruments, not app code, and both had a finding resting on them.

## 1. The journey's dictation step was failing on a spelling

P1 has been red for months. Its line was reaching the kernel every time.

The step dictates a clip that says *summarise*. iOS's recogniser writes **summaries**. The check looked for **summarize**. The run log printed the answer on the line directly above the failure:

```
P1 heard: Please summaries what the workers did today in two sentences.
P1 FAIL no /user +please summarize what the workers did/ within 30s
```

And the kernel has it — journey run 33, seq 2326. Matched now on the words the recogniser does not get to choose.

It only surfaced now because until #547 the step could not send at all: it tapped `Microphone` twice, and that button is three buttons in turn — Microphone, **Stop** while listening, then Up. Two faults stacked, the outer hiding the inner.

## 2. `kernel.py total <agent>` was answering for the root

It sent the agent name and then printed the **first** `history_end` frame it saw. Attaching starts the root's own replay, so the root's frame usually arrives first: **every worker asked about came back with the root's count**, the same number each time.

Filtering on the name alone then reported **0 for the root**, because the kernel answers `main` under `root` — the opposite mistake, made and caught in the same minute. Both halves are handled now.

Measured after:

| agent | total |
|---|---|
| `main` | 2336 |
| `say-sentence-about-mountains` | 8 |
| `say-sentence-about-rivers` | 8 |
| `writeahaikuaboutriversan` | 5 |
| a name that does not exist | 0 |

It discriminates. It did not before.

### What that unmakes

Cycle 46 concluded, from "8 of 8 workers answer `total: 0`", that empty worker chats were the kernel's truth rather than an app fault, and a features ask was filed on it. **That reading came off this tool.** The conclusion may still be right; the measurement never supported it.

What is now known, measured and on screen: a worker's chat **does** have content — opening a recent worker shows its plan, its reads and its Done line, and the kernel holds 8 lines for it.

What is not: whether an *archived* worker keeps its transcript. The original sample was `qa-cycle-11-demo`, whose tree now carries zero children, so that exact reading cannot be retaken. The question is reopened rather than answered, and I have appended a correction to the features-inbox item rather than leave a live ask standing on evidence known to be false.

## Journey run 33

First full run since the repairs of cycles 56–61, on `pod`:

| step | run 32 | run 33 |
|---|---|---|
| P2e — the photo line reaches the kernel | FAIL | **PASS** |
| P3 — the call's question lands in this project | FAIL | **PASS** |
| P2 — the model names the picture | passed on a plausible answer | **`Ice plant flowers, predominantly magenta.`** |
| P1 — the dictated line | FAIL | FAIL, on the spelling above |

Evidence and both recordings are in the project store at `media/mobile/cycle-61/run33-`.


unarbos:
Merge steward: merged — `main` is now `4d63352f`; `rust` fast-forwarded. Harness-only (two files, parse). Draft flag read as a formality. No `ios/` change, so TestFlight stays 1765.

PR #552, opened 2026-09-18
loop: kernel.py total must be told which agent a frame is for

unarbos:
Third correction to one small command tonight, and the first two were each declared fixed on a single passing run.

## The hole

#551 made `total <agent>` filter `history_end` frames by agent. It did so with a default:

```python
if f.get("type") == "history_end" and f.get("agent", agent) in want:
```

A frame with **no** `agent` field therefore defaults to the one asked about and matches. Attaching starts the root's own replay, so the root's frame is in the race — and when it arrives unlabelled it wins under any worker's name.

It did. `count-slowly-one-to-forty` read **2350**, the root's total, where its own is **11**.

The frame must say which agent it is for.

## Run three times, not once

This is what exposed it, and it is the part worth keeping:

| agent | three runs |
|---|---|
| `main` | 2350, 2350, 2350 |
| `count-slowly-one-to-forty` | **11, 11, 11** (was 2350) |
| `say-sentence-about-mountains` | 8, 8, 8 |
| a name that does not exist | 0, 0, 0 |

The earlier readings of 8, 8 and 5 were *correct* — but they were won by a race rather than by the filter, which is why one passing run said nothing about whether the command worked. A race does not care about one run.

## What this does and does not change

It does not change last cycle's conclusions: a live worker's chat has content, and the app draws it. That was confirmed on screen as well as by this command.

It does mean **any single-run reading from this command before now is worth repeating**, including the 8/8/5 above, which is why they are in the table.


unarbos:
Merge steward: merged (both commits) — `main` is now `929e681b`; `rust` fast-forwarded. Harness-only; scripts parse. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1765.

PR #554, opened 2026-09-18
loop: the journey runs the tools in the checkout, not copies in $HOME

unarbos:
The coverage row for the loop's own machine has said since cycle 41 that the harness lives in the repo, so no worker's disk is load-bearing. That was true of the **files** and not of the **paths that ran them**.

`mac-journey.sh` called four committed tools from `$HOME`:

```
7 × ~/kernel.py
1 × ~/frame-log.py
1 × ~/journey-record.py
1 × ~/push-check.sh
```

All four also exist in `deploy/mobile/`. Both copies existed, and they drifted.

## What that cost, tonight

`kernel.py` was corrected **three times** on 09-18 — the agent filter, the `main`/`root` alias, and the race on unlabelled frames. Not one of those fixes reached a journey run, because the run read the home copy.

Side by side on the same agent, just now:

```
via the checkout:  8      the worker's own count, correct
via ~/kernel.py:   2366   the root's count — the bug fixed three times
```

Journey run 33 is not invalidated by this: the journey uses `history`, not `total`. But the next fix that matters would have gone the same way, silently.

## The other half

It also removes an undocumented step from rebuilding the Mac. The recovery story written at cycle 41 — key in the vault, harness in the repo — implies a fresh machine can clone and run. It could not: it would have needed someone to know that four files must first be copied into `$HOME`, which is written down nowhere.

`~/arbos`, `~/mobile-clips` and `~/mobile-out` stay as they are. Those are the checkout, the audio fixtures and the output directory — machine paths, not tooling.


unarbos:
Merge steward: merged — `main` is now `d9a57e84`; `rust` fast-forwarded. Harness-only (one file, parses); green 5/5 on `1dadcff6`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1765.

PR #545, opened 2026-09-18
hub: a machine whose last registrant left stays on the roster, offline

unarbos:
## What

iPhone loop finding M-170: when a machine's only kernel stopped, the hub dropped the machine and `GET /list` returned `[]`. A client could not tell "this project is asleep" from "this machine was never here", which is why the phone has to remember what was opened instead of reading it from the roster.

(Was stacked on #538 for `last_activity_ms`; #538 is on `main`, so this now targets `main`.)

## How

- The hub keeps the row when the last registrant leaves. `MachineInfo` gains `online: bool` and `offline_since_ms` (Unix millis; absent while online). `online` defaults to `true` when absent, so a roster from an older hub reads as before.
- While offline, every project a kernel of that machine ever served (`MachineEntry::served`, name → place; worktrees excluded) is listed as a `live: false` row with its face, place and `last_activity_ms` kept. `since`, builds, worker projects and worktree claims are cleared — nothing live is kept.
- An attach to an offline machine is refused plainly: `mac is offline: nothing of it has been connected since <rfc3339>; it served "fieldwork" — start a kernel there and it comes back`.
- A kernel or worker registering again puts the machine online. Seven days offline forgets it (checked when the roster is built). Memory only, as the whole roster is: a hub restart forgets sooner.
- `MachineInfo::describe()` (the prompt's roster line and `machines.md`) says `offline`.
- Design doc: one line under the roster fields.

## Tests

- `arbos-hub`: `a_machine_whose_last_kernel_left_stays_on_the_roster_asleep` — row stays with `online: false`, the served project as a `live: false` row keeping identity, place and activity; JSON shape; the refusal text through `Hub::kernel` for a named and an unnamed project; forgotten after `OFFLINE_KEEP_MS`.
- `cargo test -p arbos-hub` 14 pass; core hub tests pass; desktop `cargo check` clean.

## For the phone

`machines[].online == false` is a machine asleep. Its `projects[]` rows are what it served, with `last_activity_ms` to sort by. An older hub sends no `online` field; treat absent as `true`.

[...]

PR #556, opened 2026-09-18
ios: a sleeping machine reads as asleep, not as Off

unarbos:
#545 keeps a machine on the roster after its last kernel leaves — `online: false`, with its projects `live: false` and an `offline_since_ms`. The app read neither field.

So those projects arrived looking like any idle one, and the row said **`Off`**. True, and silent about *which* thing is off — which the row's own comment already calls out as the only question worth answering there:

```swift
// The roster has stopped listing it. "Off" is true but says nothing
// about which thing is off, and that is the only question worth
// answering here: his machine, or that project's kernel.
```

The hub now answers that question and the app was not listening. Its rows name it:

```
alpha    Idle · 2m
beta     sleepy-box is asleep
gamma    sleepy-box is asleep
phone    arboslife is off
```

That last line is the existing remembered-project path (#480) — a project the roster has stopped listing at all. The two read differently on purpose: *asleep* is a machine the hub is holding open for, *off* is one it has lost track of.

## A second fault, found by trying to test the first

The live roster has no sleeping machine, and stopping somebody else's kernels to make one is not a test worth running — so the app was pointed at a small local hub serving a fixture of exactly the shape #545 documents.

It showed nothing but stale cached rows. The cause:

```swift
components.scheme = components.scheme?.lowercased() == "ws" ? "http" : "https"
```

`list()` mapped only `ws` to plain http, so an `http://` hub became `https://` and never answered. Meanwhile `attachURL()` has always used the other rule — `http` or `ws` stay plain. A plain hub on a local network was therefore **attached to over `ws` and listed from over `https`**: its roster never loaded, and the list quietly showed what it had cached, with no error.

`list()` follows `attachURL()`'s rule now. The fixture test that exposed it passes over plain `http://`, which is the same test that failed silently before.

## Testing

`deploy/mobile/scenarios/sleeping-machine.sh` runs a local hub with one machine awake and one asleep for an hour, points the app at it, and reads the rows back. It tests the phone, not the hub — the fixture's shape was checked against the live roster's own fields first, which do now carry `online` and `offline_since_ms`.


unarbos:
Merge steward: merged — `main` is now `c3adbb96`; `rust` fast-forwarded. Two `ios/` files plus a rig scenario; green 5/5 on `caac85fc`. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #557, opened 2026-09-18
ios: a slow first connect keeps the words already spoken

unarbos:
The call opens the microphone before the socket, and holds captured frames until there is somewhere to send them. That hold was **two seconds**, on this reasoning:

```swift
// the newest two seconds are worth keeping, and anyone silent
// for longer than that has said nothing to lose.
```

That is sound when the wait is the *hold's*. The wait here is the **connect's**, and what gets discarded is the opening of a sentence somebody has already spoken.

Measured on 09-18 with `-micWav`, counting frames at both ends:

| connect | frames |
|---|---|
| 2633 ms | 1050 clip / 1037 sent — **13 lost** |
| 1936 ms | 900 / 900 |
| 1514 ms | 900 / 900 |

The loss was a single event at the start and constant thereafter — the hold overflowing, exactly as designed, because the connect outlasted it.

## The change

Six seconds instead of two. `24000 × 6 × 2` is **288 KB**, which is nothing set against losing a person's first words, and audio held this long is not stale: they said it, and it still has not been sent anywhere.

## After

Four runs:

| connect | frames |
|---|---|
| 2140 ms | 1000 / 1000 |
| 2096 ms | 850 / 850 |
| 1759 ms | 850 / 850 |
| 1578 ms | 850 / 850 |

No loss in any, **including the two above the old two-second boundary**, which would have lost frames before.

**What this does not show:** a connect in the 2.6 s range did not recur across those four runs, so the specific 2633 ms case is covered by the window now being three times the old one rather than by a repeat of that measurement. Saying so because the difference matters — the mechanism is understood and the boundary has moved, but I did not re-observe the exact failing case passing.

## Not in scope

Connect-time drift itself stays a named look rather than a fix here: 5773 ms cold, then 1.4–2.1 s settled, against 1017 ms recorded at cycle 52. Widening the hold means a slow connect no longer costs the user their opening words, which is the part that was hurting.


unarbos:
Merge steward: merged — `main` is now `3e36fb8e`; `rust` fast-forwarded. One constant in one `ios/` file; green 5/5 on `05124cf5`. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #555, opened 2026-09-18
kernel: claim frame — a path a person is editing is not written by an agent

unarbos:
## What

Side-panels design, handover item 6 — the **claim** half of Jacob's co-editing ruling: *a client claims a path while its editor holds unsaved edits, and the kernel refuses the agent's write to a claimed path, telling the agent in its own turn so it can say what it wanted.*

Not in this PR: the other half (`write_if_unchanged(path, expected_hash)` for project files and a `changed` frame per watched project path). `put` already does compare-and-swap for `.arbos/` files; widening it to project files is a separate decision and PR.

## How

- **Frames** (`arbos-core::wire`): `claim { path, held }` client → kernel (`path` relative to the place or absolute within it). `claimed { path, held, by, since_ms }` kernel → every client: after each change, and on attach for each path held at that moment. A path outside the place is an `error`. Readers may not claim (the existing role gate).
- **Kernel** (`hooks.rs`, `serve.rs`): `KernelHooks.claims`, keyed by the normalised absolute path (`.`/`..` folded without touching the disk — the file may not exist yet). A claim belongs to the connection that made it (`CONN_N`): another window's release changes nothing; the connection closing releases everything it held and says so.
- **Engine** (`tools/mod.rs`): `Hooks::claimed(path) -> Option<Claim>` (default none). `preflight` checks every declared write path of a file tool (`write`, `edit`, `apply_patch`, …) and bails with `claim_refusal`: *`refused: <path> is open in <who>'s editor with unsaved changes (held for 3m). Nothing was written. Say in your reply what you wanted to change there and why, and ask them to save or close the file; then make the change.`* That text is the tool's result the model reads. `bash` is not caught — its writes are not declared — and the doc says so.

## Tests

`path_claims_e2e`: outside-the-place path → error; hold → both windows hear `claimed` (absolute path, `by`, `since_ms`); the agent's `write` is refused with the text above and the file is not made; release → the next write lands; another window's release changes nothing; a late window hears the held path on attach; the holder's connection closing releases it on the other window. Clippy clean; desktop `cargo check` clean.

## For the desktop

Send `claim {path, held: true}` when a buffer becomes dirty, `held: false` on save or close. Draw `claimed` as a held mark on the file row; the agent's refusal already reaches the chat as its tool error.

[...]

PR #558, opened 2026-09-18
kernel: save frame — a person's editor saves a project file with compare-and-swap

unarbos:
## What

Side-panels design, handover item 6 — the second half of Jacob's co-editing ruling: *`write_if_unchanged(path, expected_hash)` and a frame for the save itself.* Pairs with #555 (the claim half). Together they are what the drawer's Documents tab needs to go from read-only to editing: the agent does not write over an unsaved buffer (#555), and the person does not save over an agent's edit unseen (this).

## How

- **Frames** (`arbos-core::wire`): `save { path, text, base_hash }` client → kernel; `saved { path, size, hash, by, error }` back. `base_hash` is the sha-256 the editor read the file at, or `""` for a file that must not exist yet. `hash` on the reply is the file's sha-256 now — the next save's `base_hash`, on a conflict too, so the editor re-reads without a second round-trip.
- **Kernel** (`files.rs::save`, `serve.rs`): compare-and-swap, whole file (tmp + rename). Refused, with the reason, nothing written: a path outside the place; under `.arbos/` (the store is `put`'s, which already does CAS); `.git/`; a protected file (`AGENTS.md`, hooks, rules) unless the saver's role is owner; a folder. A save that lands is broadcast to every client, so a second window reloads; a refusal goes to the asker alone. Readers may not save (the existing role gate).
- Not a `changed` frame: `changed.path` is relative to `.arbos/` and clients mirror those with `read`/`tail`; a project path there would mislead them. `saved` carries the path relative to the place.

## Tests

`save_frame_e2e`: three refusals with their words and no bytes written; a new file with `""` lands and the other window hears `saved` with the hash; a stale base is a conflict the asker alone hears, carrying the hash now; the right base lands; the agent's `write` in between makes the editor's save a conflict that says so ("changed since you read it"), and the hash on that reply lets the next save land. Clippy clean; desktop `cargo check` clean.

## For the desktop

On open: `read` the file, keep its hash. On save: `save {path, text, base_hash}`; on `saved` with no error keep the new hash; on `conflict` show it, reload from disk (the reply has the hash), let the person redo. Pair with `claim` (#555) while the buffer is dirty.


unarbos:
Merge steward: merged `main` (`3e36fb8e`) into this branch to resolve the `serve.rs` conflict with #555 — both PRs added an arm to the same per-connection frame loop, so the `Save` and `Claim` arms are now composed in the one loop with both preludes kept. `save_frame_e2e`, `path_claims_e2e`, `protected_files_e2e` and core `contracts` pass here; desktop checks. I take it on green.

unarbos:
Merge steward: merged — `main` is now `64152bc7`; `rust` fast-forwarded. Read as a write on a client's behalf: confined to the place; `.arbos/` and `.git/` refused; protected files owner-only; folders refused; sha-256 compare-and-swap with tmp-then-rename; readers gated out. Green 5/5 on the composed head `52cb56a1`; `save_frame_e2e` and `path_claims_e2e` pass on the merged tree. Draft flag read as a formality. Mac worker: build from `64152bc7`. `v0.2.0` stays a draft.

PR #527, opened 2026-09-17
checkpoints: the refs of turns no rewind can reach are dropped — a cut's, a roll's, an archived worker's

unarbos:
## What

- A rewind's cut drops the cut turns' `refs/arbos/cp/<agent>/<line>` beside their tree sidecars.
- A transcript roll drops the rolled checkpoints' refs (read from the archived `NNNN.checkpoints.jsonl`).
- Archiving a worker drops its whole set (`git for-each-ref refs/arbos/cp/<agent>/`), before its worktree is removed.

Best effort; a ref already gone is fine. The tree commits become unreachable and git's own `gc` reclaims them in time.

## Why

Every turn that starts on a dirty tree makes a commit of the working tree and keeps it alive by a ref, and nothing ever deleted one. On a long project — the shape Jacob runs — the project's `.git` carried every turn's working tree for ever, `git gc` could reclaim none of it, and `git log --all` or any GUI that lists all refs showed thousands of `arbos-checkpoint` commits in the person's own repository. Once a checkpoint is cut, rolled, or archived, no rewind reaches it (a rewind into rolled history is refused by design), so its ref holds nothing anyone can use.

## How to check

- [x] `cargo test -p arbos-engine --lib tools::git` (18): refs present after two snapshots; one dropped by line, the kept turn's staying; a worker's set dropped by agent with another agent's untouched; a ref already gone is fine.
- [x] `cargo test -p arbos-kernel --lib rewind`; `rewind_no_identity_e2e`, `standing_pass_e2e`, `archive_children_e2e`, `history_archived_e2e`.
- [ ] `cargo test --workspace` on CI

## Notes for the reviewer

- Only refs whose checkpoint has a `work` commit are dropped; a clean turn made none.
- `checkpoint_ref` is now the one place the ref name is spelled; the writer uses it too.
- The refs are deleted, not the objects — nothing is destroyed here; git decides when to reclaim, and until then a person can still `git show` a commit they have a hash for.

[...]

PR #561, opened 2026-09-18
loop: a scenario for the dictation promise — the words wait, counted off the kernel

unarbos:
## What this is

The coverage row for voice notes carries a claim from cycle 44: dictated words sit in the composer and **nothing reaches the kernel until he presses send**. That is a correctness claim, not a cosmetic one, and until now it was checked by looking at the screen. Looking at the screen cannot see what crossed the socket.

This adds `deploy/mobile/scenarios/voice-notes-wait.sh`, which counts the kernel's own transcript at three moments: before dictation, while the words are sitting unsent in the field, and after the send.

## The measurement it produced (M-252)

```
kernel transcript before:      2411
kernel transcript, unsent:     2411   <- nothing reached the kernel while it waited
kernel transcript after send:  2415
```

The field held `Please summaries what the workers did today in two sentences.` with `Up` beside it. Same shape cycle 44 recorded on a different build (1526 → 1526 → 1531), so the promise holds and is now repeatable rather than remembered.

## The trap it avoids

The row tapped in the projects list and the kernel whose transcript is counted are **separate arguments**. They default to `phone` and `pod`, which happen to be one kernel drawn twice (M-121) — which is exactly why hardcoding the target would have gone unnoticed. Tap any other row and you must name its target, and a count that cannot be read stops the run instead of printing numbers that describe a kernel the dictation never touched.

This is the same class of fault the loop has paid for repeatedly: a rig that quietly reports something other than what it claims to measure.

## Scope

Harness only. No app code changes.


unarbos:
Merge steward: kernel red on the engine unit test `tools::git::tests::checkpoint_refs_are_dropped_by_line_and_by_agent` — second time tonight (also #555's first run), on a PR that adds one shell script. Reran; I take it on green. For #527's author: two runner reds on this test in one night, both with the first snapshot's ref absent, is a pin's worth.

unarbos:
Merge steward: merged on the rerun's green — `main` is now `c60ef366`; `rust` fast-forwarded. Harness-only (one scenario, parses). Draft flag read as a formality. No `ios/` change, so TestFlight stays 1789.

PR #562, opened 2026-09-18
voice: a pause mid-question is one question — the join window

unarbos:
## What

Inbox item `2026-09-18-one-question-two-answers-across-a-pause.md` (M-146; iPhone loop cycles 43 and 68). *"Hello Arbos. What are we working on right now?"* — a breath — *"Give me one sentence."* came back as two `transcript.final`s and two completed, spoken `response.done`s, the second over the first. The phone sent 900 of 900 frames; the split was the gateway's own segmentation: its VAD ends a segment at 600 ms of quiet and the segment was answered at once.

## The fix — a join window

`Tuning.join_ms` (default 1200; `--join-ms`). A segment that ends is transcribed at once (so Whisper's time is hidden inside the window) but **not answered until `join_ms` more quiet has passed**. Speech that resumes within the window continues the same utterance: the finished segment's words are held, and the continuation's final carries them in front. One `transcript.final`, one question to the main agent, one answer.

- Both capture paths: `pipeline.py` (`_joined`) and `duplex.py` (`_joined`), which the GPT-Live engine shares through `_uplink`/`_utterance_done`.
- Finals settle one at a time (`final_lock`), so a segment's words are held before its continuation's final looks for them, whatever order the ASR calls return in.
- The continuation decision is made from the same clock on both sides (`cap_end_at`/`last_end_at` against `join_ms`); the only disagreement possible at the boundary resolves to a merge, never to two answers.
- `speech.started`/`speech.stopped` still mark each burst as before; only the final and the answer wait.
- Latency: a reply now begins after 600 ms + 1200 ms of quiet instead of 600 ms + ASR time. Measured on the harness: the first answer ~0.6 s later than before. Being answered twice reads as the app losing track; a moment's pause reads as thinking. `--join-ms` tunes it.

Why 1200: a designed 1.0 s gap measures ~920 ms between the VAD's end and its next start (tails fade under the end threshold, onsets take `min_speech_ms` to count), so 900 sat on the edge and flaked; 1200 covers a one-second breath with margin.

## Not fixed here

In the GPT-Live engine the model also hears the raw audio and segments for itself; if it chooses to speak to the first half on its own, that is its reply, not ours, and the Live API offers no "wait for me" that I found. Our transcript, the kernel's question and the narrator's answer are one; the model's own small-talk voice may still split. Noted for the voice owner.

## Tests (`voice-server/tests`)

- Harness: a `say` step takes `gap` + `then` (one utterance with a breath in it); `asr_script` runs the gateway with `--asr mock` so **its own VAD** segments — the production shape, which the other scenarios (`--asr none`, the speech model's endpointing) never exercised; new expectations `transcript_finals`, `response_dones`, `kernel_user`.
- `pause-mid-question` and `pause-mid-question-pipeline`: one `transcript.final` with the merged text, one inbox request, the main agent asked the whole question, the answer spoken. **Control:** `python -m tests.run pause-mid-question --gateway-args --join-ms 0` fails the old way (two finals, two requests).
- All 27 scenarios pass (`python -m tests.run`).

## Reproducing on the phone

The note's own check: `xcrun simctl launch … -micWav ~/mobile-clips/pause.wav`, then `grep -E "transcript:|^event response.done"`. One of each is fixed.


unarbos:
Merge steward: merged — `main` is now `1e0e9cde`; `rust` fast-forwarded. `voice-server` only; CI 5/5 on `1eb0257c`. CI does not run the voice harness, so I ran it on the merged tree here: every scenario PASS, including the two new `pause-mid-question` ones and the existing `work-sound-activity`, `typed-during-call`, `live-session-context`. Marked ready per the coordinator. The pod's gateway runs from a branch, so it wants a redeploy from `1e0e9cde` for the phone to get the join window. `v0.2.0` stays a draft.

PR #566, opened 2026-09-18
ios: say why the link is down, once, in the phone's own words

unarbos:
## Why now

Cycle 49 established the rule — a refusal is final, a transport failure is not — but could only drive the transport half. The refusal half has sat unverified on device since (**M-165**), and the live hub cannot be made to refuse on demand.

This adds a fixture that serves the hub's own three refusals **verbatim from `Hub::kernel`**, plus a 502 at the tunnel, and counts every attach so that *retrying* and *stopping* are measured rather than guessed at.

## What the fixture found

| the hub says | before | after | attaches in 25 s |
| --- | --- | --- | --- |
| `no machine named "ghost-box" is registered` | right, but with a false line above it | `ghost-box is not connected — its kernel isn't running, or the machine is off.` | **1 — stops** |
| `fixture-box is offline: …since 2026-09-18T02:14:03Z; it served…` | the whole sentence, prefix and timestamp and remedy, in red, **once per retry** | `fixture-box has been off for 2 hours. It comes back when a kernel starts on it.` | 3 — retries |
| `fixture-box has no kernel serving "no-kernel" …` | `Link lost` — the reason was **erased a moment after arriving** | `no-kernel's kernel on fixture-box isn't running.` | 3 — retries |
| a 502 at the tunnel | correct | unchanged | 3 — retries |

## The four faults

**The reason was erased by the close that followed it.** A refusal arrives as two events — the hub's reason, then the socket going — and only the first carries words. The second, having none, overwrote the first. A person saw `Link lost` for a link the hub had just explained.

**An offline machine was not recognised at all**, so it fell through to the raw-notice branch and printed the hub's record of itself, including an RFC3339 timestamp, once for every retry attempt.

**The app guessed about the path beside an answer.** The hub sends its reason and then closes bare (`ws.close(None)`), so every refusal also looks like a transport failure. Each refused chat carried `127.0.0.1 could not be reached — retrying` above the hub's own sentence — on a host it had plainly reached and been answered by.

**An empty chat claimed to be ready** one line above the explanation of why it was not.

## A fault of mine, worth recording

The first two drafts of the fixture *paraphrased* the hub — project before machine, reason in the close frame instead of the error frame — and each draft produced a confident report of an app bug that did not exist. The app's translation was right both times. The fixture now copies `Hub::kernel` and `refuse_close` exactly, and says in its own comments why.

## Evidence

Counted, not eyeballed: the fixture logs every attach, so the difference between a socket that stopped and one that is about to come back is a number. Stills in `media/mobile/cycle-70/`.


unarbos:
Merge steward: merged — `main` is now `233ee653`; `rust` fast-forwarded. Three `ios/` files plus a fixture scenario; green 5/5 on `e8e02d35`. Draft flag read as a formality. Watching the TestFlight run for the build number.

unarbos:
Merge steward: TestFlight run for `233ee653` printed `uploaded 0.2.0 (1818)` — build 1818 carries this.

PR #570, opened 2026-09-18
prompt: the producer rule's step is reading the producer before the first edit

unarbos:
## What

Inbox item `2026-09-18-swebench-541-read-and-correction.md` from the SWE-bench loop (cycle 25, ten rollouts on the widened producer rule's two cases).

## The reading

The mark #541 added — *the root-cause line in your reply names where the fix goes* — was present in 9 of 10 rollouts, and the diff went to the producer in 3 of those 9. The loop's conclusion, which the cycle 12–24 failures bear out: the agent already knew where the fault was; the mark names the failure, it is not a lever on it. What the three producer fixes shared, as cycle 17's pairs did, is that the agent **opened the producer's code before editing**.

## The change

One clause in `CONTRACT`'s producer rule: *Before your first edit, read the function that line names — open its file at that function (the read call is the record); an edit anywhere else made without having read it is the guard at the crash site, whatever it is called.* The rest of the rule stands. A transcript shows the step as a `read` of the producer's file before the first `edit`.

The loop's correction is noted: django-13513 was the wrong evidence for the no-change rule (the request is satisfied on the base commit, and the agent said so in 5 of 5); the rule itself stands, and its real cases are django-15022 and django-10097.

## Also read, nothing to change

`2026-09-17-swebench-repro-and-instructions-gaps.md`: (1) `note_failing` taking a refused `pip download` as the reproduction — already closed on `main` (`runs_code` gate: only a command that ran code is taken); (2) the `instructions.md` shape — in the kernel only an agent def's body writes that file, at spawn; the harness was the one caller overwriting standing rules, fixed on its branch.

## Tests

`prompt.rs` pins the new clause; `budget_once_e2e` still green with the longer prompt.


unarbos:
Merge steward: kernel red on `goals_e2e::a_goal_wakes_the_agent_while_its_check_fails_and_closes_when_it_passes` (a 30 s wait for the goal's marker file). Passes 6 of 6 on `main` + this branch here, and `prompt::` 7 / `budget_once_e2e` 2 pass, so the longer prompt did not move a boundary. Reran the job; I take it on green. Another name for the timing roster.

unarbos:
Merge steward: the rerun went red on a *different* test — `arbos-update::install::tests::a_reader_watching_the_path_never_sees_it_missing` (1 miss in 4190 looks over 40 swaps). Not this PR's doing (one `prompt.rs` clause), but not a timing flake either: a reader saw the binary's path missing during a swap, which is the property the exchange exists to keep. Likely the runner's filesystem taking a non-atomic fallback. Reran again for this PR; flagging the swap finding to the update owner separately.

unarbos:
Merge steward: merged on the third attempt's green — `main` is now `0ff75662`; `rust` fast-forwarded. One clause in `CONTRACT`; `prompt::` 7 pass on the merged tree. Draft flag read as a formality. `v0.2.0` stays a draft.

PR #574, opened 2026-09-18
loop: catch the workers sheet's live row, and stop the scenario scoring on somebody else's worker

unarbos:
## The row is caught

With a `sleep` worker running, the sheet draws it with the braille spinner and its step — the row cycle 61 tried four times to catch and never did:

```
⠋, sleep 150 seconds, running sleep 150
⠙, sleep 300 seconds, sleeping 300 seconds
```

The chat's own line is right alongside it: `1 Working sleep 296 seconds · sleeping 296 seconds`.

## One open question, named rather than guessed at

Both sightings were of workers that had been running for minutes. On the single run that started from a **quiet project** and opened the sheet **18 seconds** after the send, the sheet — scrolled to its end — held twelve rows and no live one, while the chat behind it named the worker correctly. Every run, in every state, shows exactly **twelve** rows.

Three explanations fit and I have separated none of them: the sheet may lag its first publish, it may cap at twelve, or it may order a new worker outside what is rendered. The constant twelve is the thread to pull. **No app change here, because I do not yet know which it is** — and the chat's line is live and correct throughout, so nobody is left without the information.

## Two faults in the scenario, both repeats

**A first-screen count on a scrolling list.** It counted spinner rows from a single dump and reported "the sheet shows no spinner on any of them" with the live row present at y=825, below the fold. That is exactly cycle 71's mistake (M-265), two cycles later: only rendered rows reach the tree, so "it is not there" taken from one screen means "it is not on this screen". It now pages to the end and collects rows before concluding.

**It scored on the previous run's worker.** The rerun read `sleep 150 seconds` as its live row while that run had asked for 300 — the worker still going was the one started two minutes earlier. The pill also lit within four seconds, so the wait for "Working" was satisfied by somebody else's work and the sheet was read before this run's worker existed.

Fixed twice over: the nap length is randomised and serves as the run's signature, and a live row that is not this run's scores **inconclusive** rather than passing. Better than the guard, the run now waits for the project to be quiet before it sends, removing the condition the guard was compensating for.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `e7b94249`; `rust` fast-forwarded. Harness-only (one script, parses); green 5/5 on `3a6034fa`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1831.

PR #571, opened 2026-09-18
ios: the round buttons say what they are for

unarbos:
## What a person hears

With no accessibility label, SwiftUI reads the SF Symbol's own name aloud. So:

| control | announced itself as | now |
| --- | --- | --- |
| settings | `Gear Shape` | `Settings` |
| the list's filter | `PopUpButton` | `Filter` |
| the chat's overflow | `PopUpButton` | `More` |
| the project glyph in the chat header | `Move`, beside the word `pod` | hidden — it is decoration |
| back, search | already sensible | `Back`, `Search` |

A button wrapped in a `Menu` gets no name at all, because the `Menu` presents itself and the child's label never reaches the tree. Both menus needed their own.

## Two audiences, one fix

The harness has been tapping the chat's overflow **by coordinate since cycle 56**, for want of anything to ask for. It can now ask: `ui menu More`, `ui menu Filter`. The positional search stays for older builds.

## Also here: the list's oldest coverage row, measured

Search, the All/Live filter and pull-to-refresh were last exercised at cycle 47, two of the three read off a screenshot. Against a four-project fixture, counted off the accessibility tree:

```
at rest:       5 rows   alpha beta beta-two gamma pod
'beta':        2 rows   beta beta-two
'zzzz':        0 rows   and the screen says: No project matches “zzzz”.
cleared:       5 rows   alpha beta beta-two gamma pod
Live only:     3 rows   alpha beta pod
All projects:  5 rows
pull to refresh:  /list calls 2 → 3, and arrived-late reached the screen
```

The whole row holds. No app change was needed for it.

## Two faults in my own rig, both caught before filing

**A count of rendered rows is not a count of the list.** The list is a `LazyVStack`, so a row below the fold is never built and `describe-all` cannot see it. The first run read 11 rows at rest and 7 after clearing the search, and I was a step from reporting four projects that failed to come back. They were under the keyboard.

**A refresh test whose answer depended on how many times the app had asked.** The fixture added its new project from the fourth `/list` on; one run reached four and saw it, the next did not, and the second read as the app failing to refresh. The scenario now tells the fixture when to add it, and the fixture logs every `/list` — so "did the gesture make the app ask again" and "did the answer reach the screen" are two questions with two answers.

Both limits are written at the top of the scenario rather than left for the next reader.

[...]

PR #576, opened 2026-09-18
loop: the pill's count and the sheet's list, side by side

unarbos:
## What this found

Two counts of the same thing, from the same app, at the same moment:

```
the pill says:  Agents 19
the sheet lists: 12 rows, paged to the end
labels sharing a name on one screen (each would hide a row): 0
VERDICT: they disagree — pill 19, sheet 12.
```

One of the app's own two numbers is wrong.

This changes the shape of the open question from cycle 73 (M-272). The run that saw no live row eighteen seconds after a worker started may not have been watching a sheet that *lags* — the sheet is short by seven against the app's own count, and a worker that has only just started is a plausible one to be missing.

**Still not a diagnosis.** I have not read `publishWorkers` against this, and I am not changing app code on a number I cannot yet explain. The scenario exists so the question is one command rather than a reconstruction.

## Also checked: the worker chat still matches its pairing

Reopened against `cursor-reference/01-chat-view.jpg`. Plain header with a back chevron, a tick and the worker's name; collapsed tool rows with disclosure chevrons; flat surface, no bubbles; no composer, which is M-154's deliberate choice and not an omission — the tree carries no `TextField` at all. No regression, and the header now reads `Back` rather than a symbol name after #571.

## Two traps the scenario avoids, both already paid for here

It **pages to the end** of the sheet rather than reading one screen, because only rendered rows reach the tree — the mistake of cycles 71 and 73.

It counts shared labels **within a single screen**. Two workers can only hide each other where both are visible together; across ten pages every row repeats by construction, and the first version of this check duly reported twelve duplicates for a list that has none. That one is fixed in its own commit rather than quietly amended.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `7d42fa97`; `rust` fast-forwarded. Harness-only (one script, parses); green on `3a640249`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1831.

PR #578, opened 2026-09-18
loop: ask where a returning user lands, including after iOS reclaims the app

unarbos:
## The half of this row nobody had looked at

The coverage row "background minutes/hours → resume; what a returning user sees first" has been half answered since cycle 27. The resume half is done — 7 and 12 minutes backgrounded, the chat exactly as it was (M-167). The "hours" half has stood as *needs Jacob's phone*, and so was never examined.

But hours is **two** cases, and only one of them needs his phone:

- **still suspended** — iOS kept the process. That is the answered case, longer.
- **reclaimed** — iOS took the memory back, which is what actually happens overnight. Returning is then a *cold launch*, and reproducing that needs no hours at all, only the process gone.

## What it measured

```
left it at:        chat:phone

--- still suspended: Home, wait 120s, come back ---
came back to:      chat:phone   (same screen, same last three lines by md5)

--- reclaimed: the process is gone, as it would be after a night ---
came back to:      list
```

## Why there is no app change here

The app **knows** which project he was in — `settings.kernelTarget` persists, and is what the list's composer uses to name a project. So landing on the list after a cold launch is a choice the code is making, not a limit it is under.

Two answers are defensible, and I do not think the loop should pick one alone. The desktop persists and restores workspace state (`restore_panel` in `desktop/src/model/workspace.rs`), which leans toward putting him back in the chat — but I did **not** establish that it restores the *active project*, only panels and tabs within one. Changing navigation on half a comparison is exactly the kind of thing this loop has been burned by, so the question is filed for a decision instead:

`internal/features-inbox/2026-09-18-where-should-the-phone-put-a-returning-user.md`

## Scope

Harness only — one new scenario, no app code.


unarbos:
Merge steward: merged — `main` is now `e2bacd9e`; `rust` fast-forwarded. Harness-only (one script, parses); green on `4201219a`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1831.

PR #579, opened 2026-09-18
loop: check the one-wording rule instead of photographing it

unarbos:
## The rule, and why it is checkable at all

Jacob's decision at cycle 53: when Live speaks the kernel's answer, the chat keeps the **spoken** row and drops the kernel's parallel text for that turn. Never both wordings.

The two answers are written independently, so they differ — and that difference is what makes the rule testable rather than merely photographable. From this run:

| | |
| --- | --- |
| the kernel wrote | `This project (poems, sorting algorithms, and thirteen mathlib challenges…) is fully done, tested, and committed…` |
| the chat shows | `Everything's complete and committed on its branches; no work is in progress.` |

Six rows marked `Spoken`, the user's halves among them, and the kernel's wording absent from this turn. **The rule holds.**

The scenario now takes a distinctive run of the kernel's words and requires it to be absent from this turn's reply, instead of taking stills and leaving the comparison to whoever opens them.

## A fault in that check, fixed before it was believed

The first version searched the **whole screen** for the kernel's phrase, found it, and reported "both wordings are showing". It was there legitimately: spoken rows do not survive a restart, so *older* turns replay as the kernel's text — which M-179 recorded at the time.

It now reads only the rows after the last `Spoken` marker, which is this turn's reply, and prints what it found so the verdict can be audited rather than trusted. The tell was that I had already looked at the screenshot and knew what it showed; without that I would have filed a regression in a rule that was working.

## Incidentally: a third slow connect

This run connected at **2724 ms** — above M-246's 2633 ms — and both turns completed with audio played. With the journey's 2635 ms at cycle 72, two connects past that magnitude have now carried a whole conversation.

Supporting, not conclusive, and recorded that way: neither carries a frame count, and the counted rig's slowest run remains 2108 ms. Three observations point one way and none points the other.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `412012d5`; `rust` fast-forwarded. Harness-only (one script, parses); green on `9226ff1a`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1831.

PR #580, opened 2026-09-18
ios: a spent barge clip is not the same as never having one

unarbos:
## The measurement

Barge-in, measured again on current `main`:

```
event speech.started playing=true phase=speaking
metric barge_in_speech_started 181ms
metric barge_in_response_done  223ms
```

That sits inside cycle 65's pair (137/185 ms and 179/227 ms) and far below cycle 42's 500–522 ms. Connect 1888 ms, first reply audio 521 ms.

Recording: `media/mobile/cycle-77/recording-barge-in-181ms-30s.mp4`, thirty seconds — which also satisfies the every-third-cycle recording rule.

## The change

The run's tail read:

```
metric barge_in_unarmed clip=false injector=true
```

and I read it, for about a minute, as the barge having failed to arm — the "third never armed" case from M-244. It had armed, fired and been measured twenty lines earlier.

There is one barge clip per run and it is spent on the first reply, so **every later reply** prints that line. The wording gave the spent case and the never-supplied case the same name, so a successful run reported what looked like a failure.

Now:

```
metric barge_in_not_armed clip already used
metric barge_in_not_armed no clip given
metric barge_in_not_armed no injector
```

DEBUG only — no behaviour change.

## Why it is worth a commit

The fault was in the *words*, not the measurement, and the measurement was perfect. A metric that names success and failure the same way will eventually be read by someone who does not have the middle of the log in front of them — which is exactly how I read it, from the tail.


unarbos:
Merge steward: merged — `main` is now `24b56dce`; `rust` fast-forwarded. One `ios/` file, debug-only wording; green 5/5 on `83f93693`. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #582, opened 2026-09-18
loop: connect-time drift is not drift — it is two engines

unarbos:
## The isolation

"Connect-time drift" has been an open look since cycle 66, when a 2633 ms connect was found dropping the first words of a sentence. It read as drift because the numbers grew across the cycles: a few hundred milliseconds early on, then one and two seconds, then 2.6.

It needed no new experiment. Every call run the loop has ever done printed `metric connect <ms> <engine>` into a console log, and the logs are still on the Mac — 138 measurements. Grouped by engine:

```
engine      n     min median    p90    max   mean
duplex    101     284    543    732    898    526
openai     37     999   1936   2635   5773   2109

No overlap: the slowest duplex connect (898 ms) is faster
than the fastest openai one (999 ms). Two populations, not
one that drifted.
```

Of the 17 connects over two seconds, **all 17 are openai**. Of the 102 under a second, 101 are duplex.

The phone moved to GPT Live as its gateway partway through, so the engine changed underneath the measurement and the record read as one population wandering. That is why this stayed a look for thirteen cycles — nobody had grouped it.

## A second thing the grouping shows

The pre-socket audio hold is 6000 ms. The worst connect in those 138 samples is **5773 ms** — a margin of **227 ms**, under 4 per cent.

Six seconds was chosen at cycle 67 against exactly that 5773 ms reading, so this is a decision already taken rather than an oversight. But the margin is thin and was written down nowhere, and a connect slightly worse than the worst yet seen costs the caller their opening words. **I have not widened it on my own** — the scenario prints the margin on every run instead, so the next person to look does not have to derive it.

## Scope

Harness only. One script, which reads existing logs and takes no new measurements.

One fix in its own commit: the overlap test unpacked a four-tuple into two names and threw, after the table had already printed.


unarbos:
Merge steward: kernel red on a test this one-script PR cannot reach (see the run); reran. I take it on green.

unarbos:
Merge steward: merged on the rerun's green — `main` is now `43f89e7f`; `rust` fast-forwarded. Harness-only (one script, parses). Draft flag read as a formality. No `ios/` change, so TestFlight stays 1857.

PR #584, opened 2026-09-18
loop: scroll in points, not screenshot pixels — and withdraw what that cost

unarbos:
## The app was right; my scroll was broken

Cycle 74 filed a finding that the chat's pill counted 19 agents while the sheet listed 12 — the app disagreeing with itself. Re-measured with the scroll actually working:

```
the pill says:  Agents 19
the sheet lists: 19 rows, paged to the end
labels sharing a name on one screen (each would hide a row): 0
VERDICT: the two agree on 19.
```

There was never a disagreement. The kernel's tree, asked directly, holds 15 children of root with 15 distinct names, so no duplicate labels were hiding rows either.

## The mechanism, which matters more than the correction

Both scenarios I wrote this session swiped from **y=900 on a screen 852 points tall**.

Nothing errors. The gesture lands nowhere, the list never moves, and every "page" re-reads the first screen — so a list of nineteen reports as twelve, ten times over, with perfect consistency.

The screenshot is 1024 px tall and the screen is 852 pt, so any y read off a still is too large. That is exactly the pixels-versus-points trap `pt()` was written for at M-152, resurfacing in new code that did not use it.

## Fixed where it cannot recur

`page_up()` in `sim-lib.sh` computes the gesture from `SIM_PT_H` and takes only the udid, so a scenario cannot get the coordinates wrong. Both scenarios call it. Every older scenario was checked — they all stay under y=850 and are unaffected.

## What this withdraws

- **M-275 withdrawn.** The pill and sheet agree.
- **M-272 qualified to nothing measured.** Its reading ("12 rows, no live one, 18 s after the worker started") used the same broken swipe, so the sheet was never scrolled and the live row may have been below the fold throughout. Whether a brand-new worker reaches the sheet promptly is now untested rather than doubtful.

What stands is M-271: the live row exists, was caught twice, and carries its spinner and step.

Three findings rested on one broken gesture, and all three read as the app being wrong. That is the direction this loop's rig errors always point, which is worth remembering the next time a measurement makes the app look bad.


unarbos:
Merge steward: merged — `main` is now `c1607ec9`; `rust` fast-forwarded. Harness-only (three scripts, parse); green 5/5 on `6ada55b1`. Noted: M-275 withdrawn, M-272 back to unmeasured — I had passed the pill-vs-sheet disagreement upward at #576 and will correct that. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1857.

PR #586, opened 2026-09-18
loop: the list composer's four claims, measured off the tree

unarbos:
## What it measures, and what it found

The composer on the projects list carries four claims from cycle 57 (M-192). All four re-measured on current `main`, all four hold:

```
--- 1. at rest ---
  composer:        Message phone…
  and 'phone' is a row on screen

--- 2. with a filter ---
  rows left:       subnet120
  composer:        Message subnet120…
  it names one of them

--- 3. with a filter matching nothing ---
  composer:        Plan, ask, build…
  it offers no project, as it should
  and the screen says why

--- 4. above the keyboard ---
  composer top:    480 pt
  keyboard top:    568 pt
  VERDICT: the composer sits 88 pt above the keyboard
```

The 480 pt is the same number cycle 57 recorded independently, which is a pleasant check on both runs.

The fourth claim is the one a screenshot is worst at — "above the keyboard" is a comparison of two numbers, not something to eyeball — so both come from the same dump.

## Two bad reads, fixed

The keyboard check took the **first** TextField and got 177 pt. With the search open there are two, and 177 is the search box; the composer is at 490. It also looked for elements of type `Key`, which do not exist — the keyboard arrives as the `GenericElement` covering the bottom of the screen.

Both came from guessing at the tree instead of dumping it once and reading what is there. Worth noting that this failure printed **"could not read both numbers — inconclusive"** rather than a confident wrong number: same class of mistake as the off-screen swipe in #584, but caught by the code instead of by me.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `28d95798`; `rust` fast-forwarded. Harness-only (one script, parses); green on `6a4ac5f5`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1857.

PR #592, opened 2026-09-18
loop: re-check the pause split — #562 fixed half, and the breath turns out not to matter

unarbos:
## What the re-check found

The gateway has served #562 (`1e0e9cde`) since 04:13. Eleven runs in total, across two clips:

| clip | what it asks | transcripts | answers with audio |
| --- | --- | --- | --- |
| `pause.wav` | **1** question, one breath in the middle | **1**, every run (8 runs) | **2–4**, every run |
| `acceptance.wav` | **2** utterances, no pause | **2**, every run (3 runs) | **3–5**, every run |

900 of 900 frames captured and sent on all eleven, so the phone is not in it.

**#562 did its job.** The breath no longer splits the transcript — every run returns the whole question as one final. That was the first half of M-146.

**The half that remains is not a pause fault.** Transcription is correct in both clips; answers are over in both, by roughly one per question, breath or no breath. The gateway answers *any* question more than once, and the replies are distinct answers rather than one reply in parts:

```
reply: Right now, nothing's in progress; all recent workers have finished.
event response.done reason=completed playing=true
phase listening
reply: We just finished having two workers each run a timed sleep command…
```

The inbox item is re-scoped rather than left under a name that would send the next reader looking at segmentation: `internal/features-inbox/2026-09-18-one-question-two-answers-across-a-pause.md`.

## The scenario

Committed, because cycle 68 did this by hand and cycle 72 then had to rebuild the microphone rig from scratch for exactly that reason. `CLIP=<wav> SAYS=<how many things it asks> one-breath-one-answer.sh <cycle> <runs>` counts both faults separately and checks the phone's own frame counts each run.

## Two fixes to my own counter, each in its own commit

**It bundled two faults**, and printed "still splitting, every run" for runs whose transcripts were whole — reporting a gateway fix as no progress.

**It had "one question" baked in**, and called a correct two-transcript run a split. That one nearly cost the finding: without it the control run would have been called broken, and the comparison that re-scoped M-146 could not have been made. The tool would have agreed with the old theory by refusing to measure the thing that disproved it.


unarbos:
Merge steward: merged — `main` is now `17c5232e`; `rust` fast-forwarded. Harness-only (one script, parses); green on `aa2a35e8`. The re-scoped finding (multiple answers per question, independent of the pause) is passed to the voice owner. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1882.

PR #593, opened 2026-09-18
loop: page a list until it stops growing, and refuse to count one whose labels repeat

unarbos:
## The rows

Four coverage rows still dated to cycle 60, all re-measured and all holding:

| | reading |
| --- | --- |
| cold start | **3.2 s** to a list of 7 rows (3.1 s at 37, 3.0 s at 60) |
| long history | chat with composer in **1.0 s** on **2601** kernel-held lines; `Show 200 earlier lines` pages in **3.5 s** |
| background 8 s → foreground | **21** text rows before and after |
| attachments | chip with its `×` in the field, send arrow in place of the mic |

## Three faults in the harness, each one worse than the last

**Cycle 80's fix went into the two files I had open, not into the fault.** `several-workers.sh` counted the workers sheet off a *single screen* — 12 rows against a pill of 22 — which is exactly what M-287 withdrew a finding over. I fixed the two scenarios written that hour and never swept for the third, which already had it. A fix applied to the files in front of you is not a fix to the fault.

**"Paged to the end" was eight pages and a hope.** With 25 agents on the sheet, a fixed loop collected 17 rows and called that the end. A page count is an assumption about list length dressed as a measurement. `collect_rows` in `sim-lib.sh` now pages until two pages in a row add nothing and **reports whether it converged or hit its ceiling**, so a finished count can be told from an unfinished one.

**And then, with paging provably converged, the numbers still disagreed — and the instrument was still the reason.** Pill 25, label count 17, and the duplicate check reporting **3 labels shared on a single screen**. Rows are collected by goal text, so repeated goals count once, and this project has run enough poems and sleeps to have them.

The comparison now **refuses**:

```
VERDICT: cannot say. 3 label(s) are shared on a single screen, and rows are
         counted by label, so the sheet's 17 is a floor and not a count.
         Pill 25. Compare these two only on a project whose goals are distinct.
```

Cycle 80's clean reading stands as the real one: pill 19, sheet 19, zero shared labels.

That is the third time this session a number has looked like the app disagreeing with itself and turned out to be the instrument. The difference today is that the instrument said so itself, rather than my noticing afterwards.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `c9e035c9` (#593 and #595 together); `rust` fast-forwarded. Harness-only; scripts parse; green 5/5. Draft flag read as a formality.

PR #595, opened 2026-09-18
loop: read the settings sheet where its contents actually are

unarbos:
## The row holds

The settings sheet, re-measured on current `main`:

- five sections — `Settings`, `Voice`, `Arbos kernel`, `Mesh hub`, `Notifications`
- three saved-token fields reading `Token saved · paste to replace`
- at the foot, `Arbos, 0.2.0 (1)` under `The TestFlight build on this phone.`
- a token that cannot work drops the list to **1 row** (the pod, which needs no hub) and the screen says **`Hub token refused.`**
- restoring the token brings **7** rows back

That is M-206's visible refusal and M-201's themed ground both still standing, 27 cycles on.

## Naming the buttons broke a script that depended on the bad name

#571 renamed the settings disc from SwiftUI's `Gear Shape` — the SF Symbol read aloud — to `Settings`. This scenario tapped `Gear Shape`, found nothing, printed `no settings button` and exited. On a build that had been improved.

Fixed with the name first and the old rendering as a fallback, so it runs either side of the change. Worth recording as a cost of that work rather than a surprise: **accessibility labels are an interface**, and the harness was quietly coupled to their absence. Anything reading a symbol-derived name is reading a bug.

## And a fourth script reading one screen of a scrolling view

The build-line check read the first screen of the sheet, found nothing, and printed an empty build line — as though the app had lost it. It is four pages down, exactly where M-220 recorded it.

This is the fourth tonight after M-287, M-303 and M-304, and the first where the missing thing was **text** rather than rows. That is why the earlier fixes did not catch it: they were all about counting. The rule is wider than counting — *anything* read off a scrolling view has to be read after moving to it.

## Scope

Harness only.


unarbos:
Merge steward: merged — `main` is now `c9e035c9`; `rust` fast-forwarded. Harness-only; scripts parse; green 5/5. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1882.

PR #596, opened 2026-09-18
dev channel: wait for a newer commit only while it looks like finishing

unarbos:
Stepping aside for a commit about to supersede this one saves a runner and an Apple notarisation. Stepping aside for one nowhere near finishing just starves the channel.

## What it cost

On the morning of 2026-09-18, **twelve merges an hour** outran CI continuously. Every dev-channel run deferred, and **every deferral was correct** — I read the reasoning of nine in a row:

```
06:36  24b56dc is newer and still running…
06:33  412012d is newer and still running…
06:28  412012d is newer and still running…
06:26  8d42d33 is newer and still running…
```

Nine of the last ten CI runs on `main` were green. Nothing was broken anywhere. And the feed sat **eighty minutes and forty-two builds** behind `main`, moving only when merging happened to pause. An update button that does not move is the thing this channel exists to prevent.

## The change

`api_state` now reports *when* the oldest unfinished run started, and a pending commit holds the channel only while it has been running less than `PATIENCE` (600s). Past that the newest green commit goes out; if the pending one turns green later it publishes then and supersedes — one extra build, against a channel that keeps up.

A timestamp that will not parse counts as long ago. Something we cannot read is not a reason to hold the channel.

## Tests

The harness learned a second kind of pending — one started thirty seconds ago, one started half an hour ago — so the two can be told apart:

```
  ok    a newer commit that just started is still worth waiting for
  ok    a newer commit that has run past the patience does not hold it
  ok    one stuck commit does not hide a fresher one behind it
```

Checked by making the wait unconditional again, where the second fails with `expected 3, decided none`. The third passes either way **by design**: it pins that a fresh pending behind a stuck one still holds the channel, which is true under both rules — worth saying so rather than letting it look like three discriminating cases when only one is.

Fifteen cases pass in total.


unarbos:
Merge steward: merged — `main` is now `883a47da`; `rust` fast-forwarded. Read on the publish path: no permission or secret change; an unparseable start time is treated as long ago (the `echo 999999` after the early return is unreachable, harmless). Hermetic test 15 of 15 on the merged tree. Draft flag read as a formality. This morning's eighty-minute feed lag was exactly this; watching the next runs. `v0.2.0` stays a draft.

PR #590, opened 2026-09-18
ios: the call screen tells VoiceOver what it is doing

unarbos:
## What was wrong

`CallView` carried **no accessibility label anywhere**.

The screen is wordless during a call by design — `hint` returns nil for connecting, listening, thinking and speaking, with the comment *"In a call the screen stays wordless"* — and the orb's **colour** carries the state instead. That is a real design decision and this change does not touch it.

But colour carries to one kind of person. Read aloud, this screen said nothing whatever about whether it was listening, thinking or speaking. The three discs were unnamed too, so they read as their SF Symbols.

## What it does now

The orb is a button labelled `Call` whose **value** is the phase, taken from the `Phase.label` that already existed:

```
197  409  Button       Call        (value: listening / thinking / speaking / call failed)
 40   81  PopUpButton  Call menu
353   81  Button       Settings
```

`CallDisc` gained a label the same way `RoundButton` did in #571, so the discs read `Call menu`, `Settings`, and `Mute`/`Unmute` — the mute disc says which of the two it will do, since its symbol is the only other clue. The close button reads `End call`.

**No text is drawn.** The wordless design is exactly as it was.

## How it was found

A video review of the barge-in recording was asked whether the clip demonstrates an interruption. It reported that the orb changes colour, there are no captions or status words, and a viewer cannot tell that a reply was cut off — and asked what someone who cannot hear the call would have.

That question was about the recording. The answer turned out to be about the app.

## Also recorded, not in this diff

The same review means my claim that the recording "demonstrates barge-in" was too strong. The counts are the evidence — 181 ms to speech-started, 223 ms to response-done — and the recording illustrates very little, because the thing being demonstrated is inaudible and unwritten. Noted as M-296 rather than quietly dropped.

[...]

PR #599, opened 2026-09-18
ios+loop: name the composer's send and microphone, and move every script off the symbol names

unarbos:
## The row

The call pulled down, all four claims measured on current `main`:

- the row reads `Add`, `Type to phone`, `Mute`, `End call`
- a line typed on the call reaches the kernel — seq 2745, `heard`
- tapping `Mute` turns the disc to **`Unmute`**
- closing returns to the chat the call was entered from

The mute half had never actually been exercised: the scenario was tapping **the composer's** microphone, which lives on a different screen, and reporting "no mic button to mute" from the one screen that has a mute.

## The sweep

#571 broke the settings scenario (M-307), and #590 broke this one. Rather than patch one more file, I compared **every name the scripts tap** against the labels the app actually defines.

Four scripts still tapped `Close` for a button now called `End call` — including **`mac-journey.sh`**, where the fallback is a coordinate tap that lands on **Back**. The acceptance journey has been closing calls by hitting the wrong control.

Moved together in one commit: four `Close` taps, ten `Up` taps, and the app gaining real labels for the composer's send and microphone buttons.

## The underlying rule, now written down

The harness had been tapping `Up`, `Microphone`, `Gear Shape`, `Close` — **none of which the app ever said**. They are what SwiftUI reads out when there is no label. Every one was a defect the scripts had come to depend on, which is why each accessibility fix broke a script.

> A name a script taps should be one the app states in `accessibilityLabel`. If it is not, the script is coupled to a *missing* label, and it will break the moment somebody supplies one.

Both sides now move in the same commit.


unarbos:
Merge steward: merged — `main` is now `54baaceb`; `rust` fast-forwarded. One `ios/` file plus ten scripts moved onto stated labels; iPhone green; scripts parse. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #600, opened 2026-09-18
loop: a worker's chat, reached from the sheet — and a claim narrowed rather than a bug filed

unarbos:
## What holds

From the project chat: the pill opens the sheet, the first row opens that worker's chat with its name in the header, the chat holds its content and **no composer** (`TextField` count 0 — M-154's deliberate choice), and `Back` returns to the **project chat** rather than the sheet or the list.

## A claim that needed narrowing, not a bug

The row has claimed two ways into a worker's chat since cycle 32. Tapping `w074832 mountains · Turn ended. Last words: …` does nothing, and I was one line from filing that.

`workerLines` in `MainChatView` draws a `Button` **per running worker and nothing for the rest**. The `Turn ended` line is kernel transcript text, not a control — by design. The second way in exists only *while a worker runs*, and the row had been implying every worker line is a door.

## What is still untested, said plainly

With the target corrected the scenario starts a sleeper and waits for its line. None appeared in 80 seconds, and the likely reason is the scenario's own: it types and sends **without the read-back** every other typed line in this harness uses (M-180), so a composer that was not focused sends nothing and the run cannot tell.

Left open deliberately. Two of the row's three claims are measured; the third is honestly unmeasured rather than quietly dropped, and the scenario is committed with the right target so the next cycle finishes it.

## A scroll in the wrong direction looks exactly like an absence

The first search for a worker's line used `page_up`, which drags content upward and walks toward the **newest** line. The line it wanted was older. It reported "no Done line in view" and would have done so for ever.

`page_back` walks the other way. Worth keeping beside M-287: a gesture that does nothing and a gesture that goes the wrong way both read as *the thing is not there*.


unarbos:
Merge steward: merged — `main` is now `e70e1b5c` (#600 and #601 together); `rust` fast-forwarded. Harness-only; scripts parse; green 5/5. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1904.

PR #604, opened 2026-09-18
dev channel: the deferral clock runs on the green build, not the pending one

unarbos:
## The cap that never fires

#596 bounded the deferral: a commit still under test holds the channel only while it has been running less than `PATIENCE` (600 s). The bound is read off the **pending** commit, and that is the flaw. While `main` merges faster than CI finishes, the newest commit under test is always a young one — so it is always worth waiting for, the cap never fires, and the chain does not end.

It went on happening after #596 merged at 08:01 this morning:

| run | deferred to | age of that commit |
| --- | --- | --- |
| 08:02 | `11f9f6c` | ~3 min |
| 08:10 | `54baace` | ~2 min |
| 08:17 | `54baace` | ~2 min |

Every one of those deferrals is correct on its own terms. Together they are a feed that sits behind `main` for as long as merging continues, which is the thing this channel exists to prevent.

## The clock moves to the green build

The walk now carries on past the newest green, down to the build the channel already has. The greens behind the newest one are builds that *could* have gone out and did not, and the oldest of them says how long the channel has stood still.

Past `PATIENCE`, the newest green publishes whatever is still running, and the pending commit supersedes it when it lands. No merge can reset that question — which is the entire difference from the pending-side clock.

Three supporting changes:

- `api_state` reports *when* a commit went green (`success <timestamp>`), not only that it did.
- The triggering commit is now asked of the API so its green timestamp is real, but a `pending` answer from its sibling run is still discarded. A run still cannot wait on itself — the case #517 fixed.
- The feed is read **before** the walk rather than after it, so the walk knows where to stop. A green already on the channel has not been kept waiting and must not start the clock.

## What is tested

`.github/dev-channel-plan-test.sh` gains five cases. Two of them fail against the logic this replaces:

```
FAIL  a green that has waited goes out, whatever is still running — expected 3, decided none
FAIL  the green that has waited starts the clock; the newest green goes — expected 3, decided none
```

The other three guard this change rather than the old one: a green already on the channel does not start the clock; nothing newer than the channel being green publishes nothing; and a commit still running *behind* the newest green holds nothing — that last one matters because reading the clock is what first brings those older commits into view.

All eighteen cases pass with the change.

## It now runs in CI

New `publishing` job in `ci.yml`. The plan step decides which build reaches Jacob's Update button; it has been wrong twice, and both times it was a shell step that nothing ran except the channel itself. The test extracts the step from the workflow, decides real cases against a throwaway git repository, and reaches nothing — six seconds on `ubuntu-latest`.

It earned its place on the first run. The harness set the tip with `git update-ref FETCH_HEAD`, which git 2.49 and later refuse as a pseudoref: it passed on this machine's git 2.43 and failed on the runner's 2.55, with all twelve cases deciding "none" for a reason that had nothing to do with any decision under test. The second commit here writes the `FETCH_HEAD` file the way a real fetch leaves it, and the harness now checks that footing before it decides anything, so a stand-in that stops working says one clear line instead of twelve misleading failures.

## Cost

A build. When the cap fires, the channel publishes a green that a newer commit supersedes minutes later. That is the trade named in #596 and it is the right way round: a wasted runner against an Update button that does not move.


unarbos:
Merge steward: merged — `main` is now `0c641fff`; `rust` fast-forwarded. Read on the publish path: the clock now runs on the oldest green above the feed's build, so a stream of young pending commits cannot reset it; the triggering commit is asked of the API for a real timestamp but its sibling's `pending` is still discarded; the new `publishing` job adds no permissions. Hermetic test 18 of 18 on the merged tree. Watching the next runs for the first `Waited long enough`. `v0.2.0` stays a draft.

PR #605, opened 2026-09-18
loop: every script runs the tool beside it, and a check that keeps it that way

unarbos:
## The drift was real, and large

M-238 fixed the journey at cycle 63 to run tools from the checkout. It fixed nothing else. **Eleven more scripts** still called `~/kernel.py` and `~/find_row.py`.

Measured on the Mac, home copies against the repository's:

| tool | home | repo | |
| --- | --- | --- | --- |
| `ui.py` | 3480 b | 8310 b | less than half the current file |
| `kernel.py` | 7404 b | 8432 b | missing the agent filter from M-224 |
| `find_row.py` | same | same | retired, never changed |

The `ui.py` those scripts were using predates `field`, `focus`, `plain`, `on_screen` and `menu` — everything cycles 54 to 56 built to make this harness reliable. They were running five-week-old tooling while the fixes sat in the repository, committed and unused.

This is M-133's failure in a worse form than the original. The first time, the tooling existed only on one machine and losing the machine nearly lost it. This time the repository had it all along and the scripts did not look there — which is harder to notice, because nothing ever breaks loudly.

## The change

All eleven now resolve tools through `$HERE`, so the tool beside the script is the tool that runs.

And `deploy/mobile/check-tools.sh` fails if any script reaches outside the checkout for a `.py` or `.sh`, and prints the checksum of each tool the harness ships:

```
scripts under .../deploy/mobile reaching outside the checkout for a tool:
  none — every tool resolves inside the repository

tools the harness ships:
  kernel.py            d054c334
  ui.py                af85c626
  journey-record.py    2685c691
  find_row.py          64e46575
  sim-lib.sh           fff90387
  mirror-docs.sh       b5075571
```

Identical on both sides.

M-238 recommended exactly this check and nobody wrote it, which is why the same fault survived thirty cycles in eleven other files. A recommendation in a ledger is not a check.


unarbos:
Merge steward: merged — `main` is now `b3770cd0`; `rust` fast-forwarded. Harness-only (twelve scripts, parse; `check-tools.sh` reports none reaching outside the checkout here); green 5/5 on `c8ac7242`. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1904.

PR #607, opened 2026-09-18
loop: read the orb's phases — a row that was unmeasurable until #590

unarbos:
## A measurement that did not exist a few cycles ago

The "call — voice first, orb, colours" row has never had a check of its own, only timings borrowed from journey runs. The reason is simple: the orb's state existed **only as a colour**. Nothing in the accessibility tree said listening, thinking or speaking, so there was nothing to read.

#590 gave the orb a value carrying `Phase.label`, for VoiceOver's sake. The side effect is that the sequence a caller experiences is now machine-readable.

One question, one call:

```
time  phase
  3.2  connecting
  5.8  listening
  8.4  thinking
 20.1  speaking
 26.6  listening

VERDICT: one pass through the phases per question, in order, no flapping
```

One pass, in order, no flapping — and one spoken answer for one question, which is M-146 closed again from a different direction.

Worth noting where the instrument came from: an accessibility fix made for someone who cannot see the colour, which handed the loop a way to measure something it had never been able to measure.

## "One question" baked in, for the third time

Run against `acceptance.wav`, which asks two things, the flapping check reported `re-entered listening, speaking, thinking — check for flapping`. Each turn legitimately walks thinking → speaking → listening, so a two-utterance clip revisits all three. An ordinary two-turn call was called a fault.

Fixed the same way as M-301 and M-311 before it: `SAYS` says how many things the clip asks, and spoken answers are compared against it.

Three times now I have written a checker that assumes one turn. The assumption is invisible while the default clip has one question in it, which is exactly why it keeps surviving.


unarbos:
Merge steward: merged — `main` is now `4e59cb70`; `rust` fast-forwarded. Harness-only (one scenario, parses); green 5/5. Draft flag read as a formality. No `ios/` change, so TestFlight stays 1904.

PR #608, opened 2026-09-18
prompt: when you can name the wrong predicate, change the predicate

unarbos:
## What

Inbox item `2026-09-18-swebench-producer-step-read.md` (SWE-bench cycle 27, ten rollouts on the two producer cases, kernel `17c5232e`).

## What the read said

- **xarray-6938** (a method returns `self` where a copy is needed): 5 of 5 read `to_index_variable` before the first edit, 5 of 5 fixed it there, 5 of 5 solved. Cycle 22: 0 of 3 at the producer; cycle 25: 2 of 5. The reading step (#570) works where the fix is visible in the producer.
- **sympy-17318** (a wrong condition in `_sqrt_match`): 4 of 5 read the producer and named the condition as the fault — and then guarded two downstream callers anyway. All four fail the hidden test; the one that changed the predicate passed. Reading decides the first kind; it does not decide a judgement.

## The change

One sentence in the producer rule, where the loop pointed: *When you can name the wrong predicate, change the predicate: a guard that lets the wrong value reach a different caller is not the conservative choice, it is the same bug with one caller patched* — with the cycle's case as the example. Pinned in `prompt.rs`; `budget_once_e2e` still green with the longer prompt.

The loop also noted the `"producer: …"` reply line appeared in none of the ten; the read call is the mark that shows. The line is left in the rule for now — it costs nothing and may yet appear — but if a later read still finds zero, dropping it is a one-line follow-up.


unarbos:
Merge steward: merged — `main` is now `7bf144ca`; `rust` fast-forwarded. One clause in `CONTRACT`; `prompt::` pass on the merged tree. Draft flag read as a formality. `v0.2.0` stays a draft.

PR #612, opened 2026-09-18
loop: the running-worker line was there all along, and it opens — four rig faults, none of them the app

unarbos:
## The row is closed, and the app was right at every step

Tapping a running worker's line **by its frame** opens that worker's chat first time: `Back`, the spinner, and `q092440 one` in the header. The presentation is a `navigationDestination`, not a sheet, so the screen genuinely changes. Both ways into a worker's chat are now verified, and `Back` returns to the project chat.

## Four faults, four cycles, one small row

The running-worker half took cycles 90, 91, 98 and 99:

1. **a send with no read-back** — no worker started and the run could not tell a failed app from keystrokes that went nowhere;
2. **a scroll in the wrong direction** — `page_up` walks toward the newest line while the thing sought was older, so it reported an absence for ever;
3. **a pattern blind to the animated spinner** — the label is `⠙, 2 Working <name> · Running sleep 80` and the pattern wanted a digit straight after `Button`; plus a sibling of the same fault in the pill, plus a child-check hunting for the sleep duration in a name derived from the *goal*, which printed the false line `the root answered it itself`;
4. **a tap by label that cannot be matched** — the spinner makes an exact label stale the moment it is read, and the worker's name appears on more than one element, so a name match can land somewhere that is not the button.

**Not one was the app.** The line, the pill, the naming and the navigation were correct throughout.

## What that is worth

Four independent mistakes in one small area all produced the same shape of output — *the app did not do the thing* — because that is what a rig failure looks like from outside.

A negative result about the app deserves more scepticism than a positive one. After tonight this loop has the tally to say that rather than just the instinct.


unarbos:
Merge steward: merged on the rerun's green — `main` is now `f8b4ef27`; `rust` fast-forwarded. The first red was `the_coordinators_tool_list_says_spawn_first…`, a kernel test a one-script PR cannot reach (it reddened `main` yesterday too — one more for the roster). Harness-only; parses. No `ios/` change, so TestFlight stays 1936.

PR #614, opened 2026-09-18
dev channel test: the queue of young commits, which is the shape that starved it

unarbos:
## What this does not change

Nothing in the workflow. The rule asked for — *time the newest unpublished green, and publish it past ten minutes even when a newer commit is still running* — is what #604 already does, and #604 is on `main` as `0c641ff`, unmodified by the steward.

The walk goes newest to oldest. The **first** green it meets is the newest unpublished one, and its age is the first thing checked:

```
success*)
  : "${choose:=$candidate}"
  ready="$(seconds_since "${state#success }")"
  if [ "${ready:-999999}" -ge "$PATIENCE" ]; then
    waited="$candidate"
    break
  fi
```

So a newest green that has sat past `PATIENCE` sets `waited` immediately, and the deferral does not happen. #604 goes one step further than the rule as stated: it keeps walking to the greens *behind* the newest one, so an older unpublished green that has been kept waiting also releases the channel. That is strictly more eager, never less, and it is deliberate — the question is whether anything publishable has been sitting, not only the most recent thing.

## What this adds

One case, because the existing ones miss the shape that actually starved the channel.

Every case so far has a *single* newer commit under test. The real morning had a queue: commit after commit, each genuinely young, and by the time one finished the next was younger still. That is why the pending-side cap could never fire — there was always a fresh commit at the tip to defer to.

```
check "a queue of young commits does not outlast a green that has waited" \
  "${SHA[4]}" "${SHA[2]}" 0 2 \
  "${SHA[4]}" pending "${SHA[3]}" pending "${SHA[2]}" waited
```

Two pending commits ahead of a green that has sat. On `main` today it decides **publish build 2**. Against the pending-side cap (#596, `883a47d`) it decides nothing:

```
FAIL  a green that has waited goes out, whatever is still running — expected 3, decided none
FAIL  a queue of young commits does not outlast a green that has waited — expected 2, decided none
FAIL  the green that has waited starts the clock; the newest green goes — expected 3, decided none
3 failed
```

So the new case has the same teeth as the two #604 added, and it is the only one of the three in the shape the incident took.

All nineteen cases pass on `main` with this added. The `publishing` job in CI runs them.


unarbos:
Merge steward: merged — `main` is now `ad742379`; `rust` fast-forwarded. Test-only; hermetic test 19 of 19 on the merged tree. `v0.2.0` stays a draft.

PR #615, opened 2026-09-18
ios: a cold start comes back to the chat that was in front

unarbos:
## The decision, and where it came from

Cycle 75 measured that after iOS reclaims the app — what actually happens overnight — the phone came back to the **projects list**, not the chat the person left. I filed the question rather than choosing, because it needed the desktop's answer.

The features agent supplied it from the desktop's own code (`internal/features-inbox/2026-09-18-returning-user-the-desktop-reopens-where-you-were.md`): `state.toml` keeps the open projects **and** which was `active`, and `Workspace::new` makes that one front again. The list is a picker you open, not the landing screen.

## What changed

`kernelTarget` could not answer "which was in front" on its own: it names the last project opened and stays set while the list is showing, because the list's composer talks to it. So this adds `frontProject`, set while a chat is on the navigation stack and cleared when it is not — the same distinction the desktop draws between its open tabs and its active one.

Restored **once** at launch, not on every appearance, so leaving the chat does not put you straight back into it.

## Verified with the scenario that found the problem

```
left it at:        chat:phone

--- still suspended: Home, wait 60s, come back ---
came back to:      chat:phone   (same screen, same last three lines)

--- reclaimed: the process is gone, as it would be after a night ---
came back to:      chat:phone
VERDICT: put back in the chat he left, even though the app had been killed
```

## Two checks had to follow the feature

The scenario read the restored chat's title at 9 s, before the project's identity arrived, and reported a **different project**. And on the next run it could not find the `phone` row at all, because the app now opens into a chat rather than the list.

Both fixed — wait for the name to settle, and step back to the list before starting. Ordinary cost of shipping behaviour: a change to where the app lands invalidates every scenario that assumed the old landing, and the first symptom was a false "somewhere else entirely".

The stale-link worry from the original note is already handled: cycle 70's work means a restored chat whose link is down says so plainly.


unarbos:
Merge steward: kernel red on a test this `ios/`+harness PR cannot reach (see the run); reran. I take it on green.

unarbos:
Merge steward: merged on the rerun's green — `main` is now `9dc5d6ff`; `rust` fast-forwarded. Two `ios/` files plus the scenario; iPhone green. Draft flag read as a formality. Watching the TestFlight run for the build number.

PR #617, opened 2026-09-18
loop: clear the trigger a previous run left behind

unarbos:
## The row holds

The list's search, filter and refresh, thirty cycles after they were last measured:

```
at rest:            5 rows
typing 'beta':      2 rows — beta beta-two
typing 'zzzz':      0 rows, and: No project matches “zzzz”.
cleared:            5 rows
Live only:          3 rows — alpha beta pod
All projects:       5 rows
pull to refresh:    /list 2 → 3 calls, and arrived-late reached the screen
```

Same numbers as cycle 71, on a build many changes later.

## The fault

The fixture serves its "new" project when `/tmp/fixture-add-late` exists — and a file outlives the run that made it. Left behind by cycle 71, it served that project from the **first** call, so `arrived-late` was on screen *before* the pull and the refresh verdict congratulated itself on a row that had always been there.

The refetch count was real. The "new project appeared" half was not.

Removed at start now, and if the trigger is somehow present before the pull the run says plainly that half means nothing.

## Why it is worth recording

Same family as M-274, where a worker left running by the previous run made its evidence look like this one's. State on the machine outliving the run is its own class of fault — and both times it produced a **pass**, which is the harder direction to notice. Every other rig fault tonight made the app look broken; these two made it look fine.


unarbos:
Merge steward: merged — `main` is now `d19b2c82`; `rust` fast-forwarded. Harness-only (one script, parses); green 6/6. Draft flag read as a formality.

PR #618, opened 2026-09-18
loop: a check built on a coincidence, and the coincidence ended

unarbos:
## The rule holds; the check that tested it did not

Jacob's rule is that a delegated turn shows the **spoken** wording only, never both. The run first reported `both wordings are showing`. It had not — the chat showed the answer **once**.

The check asked whether this turn's reply *differed from* the kernel's wording. That works only while the two differ. Since today's gateway change Live often says **exactly** what the kernel said, so the single spoken row matched the phrase and was counted as a second copy.

## What M-279 encoded without meaning to

M-279 made the rule testable by noticing that the two answers differ, and said so explicitly: *"the two answers are written independently, so they differ — and that difference is what makes the rule testable."*

True when written. It is also a fact about the gateway, and the gateway moved. The check had no way to notice; it reported a violation of a rule that was being kept.

## Replaced by counting

Count the rows this turn gives its answer. No assumption about the words:

```
rows marked Spoken: 4; in this turn the answer appears 1 time(s)
VERDICT: this turn has one row for its answer — the rule holds
(0 means the spoken wording differs from the kernel's; 1 means they
 coincide, which the gateway now often does. Either is one row.)
```

Counting the *whole screen* then over-counted, because older turns replay as the kernel's text once their spoken rows are gone — so it is scoped to the rows after the last `Spoken` marker.

## Why this one is worth reading

Two corrections in one cycle, both from the same root: **the world moved under a check that had encoded a fact about it.**

That is a different failure from the rig faults earlier today, and probably more dangerous. The check was correct when written, correct when reviewed, correct when used — and became wrong without anyone touching it.


unarbos:
Merge steward: merged — `main` is now `6fe1e070`; `rust` fast-forwarded. Harness-only; parses; green 6/6. Draft flag read as a formality.

PR #619, opened 2026-09-18
loop: run the scenarios together — and it found a rotted check on its first run

unarbos:
## Why a sweep

Today the loop found, one cycle at a time, that its own checks rot: a gesture that stopped reaching the screen, a page count mistaken for the end of a list, a counting rule invalidated by its data, a pattern blind to an animated glyph, a tap on a name the app stopped using, and a check whose premise was a fact about the gateway that later changed.

**Every one was found by accident, while looking at something else.** A scenario that no longer measures what it claims goes on passing quietly. So: run them together and read the verdicts side by side.

## What the first run said

```
list-composer                 the composer sits 88 pt above the keyboard
list-search-filter-refresh    the pull refetched and the new project reached the screen
cold-start-and-history        (no verdict) composer row: …
settings-and-a-bad-token      (no verdict) 09:54:58 shot 05-rows-restored
chip-and-send-arrow           the chip goes, and the microphone comes back with it
voice-notes-wait              waited, then went — 2898 → 2898 unsent → 2900 on the send
the-core-chat-path            (no verdict) 2907 turn_complete
call-pulled-down              (no verdict) close landed on: Button Back
worker-chat-open-and-back     a running worker's line opens its chat, as the sheet does
pill-count-vs-sheet           they disagree — pill 2, sheet 3, and no shared labels to explain it.
```

## The rotted one

`pill-count-vs-sheet` is wrong, and the project was behaving exactly as designed.

The pill shows **`Agents N`** — every agent — when idle, and **`Working N`** — only those running — when busy. The sheet always lists them all. Comparing against the `Working` form compares a subset with a whole.

That is the **third** assumption this one tool had encoded, after the fixed page count and the count-by-label, and the **fourth** time it has accused the app. It now declines unless the pill is in its `Agents` form.

## Two other things the sweep surfaced

**The worker-chat row passes unattended now** — cycle 99 proved it by hand with a frame tap; this is the committed scenario reaching the same conclusion on its own.

**Four scenarios reach no verdict at all.** They print measurements and stop, so the sweep fell back to their last line of output. Not broken — but a scenario that measures without concluding is half a check, and it is invisible until something tries to read them in bulk. Named, not yet fixed.


unarbos:
Merge steward: merged — `main` is now `d12118e6`; `rust` fast-forwarded. Harness-only (`sweep.sh` new, one scenario corrected; parse); green. Draft flag read as a formality.

PR #601, opened 2026-09-18
prompt: a request that quotes its reference fixes what done means

unarbos:
## What

The G-override candidate from the SWE-bench loop's cycle 26 (`internal/qa/inbox/2026-09-18-swebench-loop-cycle-26.md`): django-10097 quotes RFC 1738 §3.1 and asks `URLValidator` to reject `:`, `@` and `/` in the user and password fields. Three of five rollouts on the #541 kernel permitted `:` in the password; two argued the wider standard in the reply. The loop's read: *the agent's reading over the request* — four cases on this instance now. The no-change rule reaches a decline; this shape is an override, and nothing in `CONTRACT` named it.

## The clause

One sentence in the done-criterion paragraph, beside the existing "the reference is the one the request says is right":

> A request that quotes its reference (an RFC line, a documented rule, a maintainer's words) fixes what done means: implement the quote as written, even where your own reading of the wider standard would allow more (…); the disagreement is one line in the reply — "as quoted: <the rule>; a wider reading would allow <what>" — never a looser check in the code.

Visible mark: the `as quoted:` line, greppable in a transcript when the agent disagreed and still did as asked.

## Tests

`prompt.rs` pins the clause and its mark; `budget_once_e2e` still green with the longer prompt.

Also this turn, no PR: the phone's returning-user question answered from the desktop's code and a cold launch (`internal/features-inbox/2026-09-18-returning-user-the-desktop-reopens-where-you-were.md`).

<div><a href="https://cursor.com/agents/bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #627, opened 2026-09-18
prompt: the quoted-reference rule's mark is a test that asserts the quote, not a reply line (SWE-bench cycle 28)

unarbos:
## What

One sentence of `CONTRACT` reworded — the quoted-reference rule from #601. The mark moves out of the reply and into the work:

> before the edit, write the quote as the assertion of a test — the test is the record that you took the quote as the rule — then implement what that test asserts, even where your own reading of the wider standard would allow more … a wider reading is one line in the reply at most, never a looser check in the code and never a looser test.

The `"as quoted: <the rule>; a wider reading would allow <what>"` reply line is gone. The example's count now reads seven of twelve across four cycles. The pin test follows the new words; its doubled `#[test]` attribute (a clippy `duplicated attribute` warning on `main`) goes.

## Why

`internal/features-inbox/2026-09-18-swebench-601-read-10097-ungradeable.md`: five rollouts on django-10097 with #601 in the contract gave the same two-forbid / three-permit split as cycle 26 without it, and **none of five wrote the reply line**. The loop's standing finding holds again — the reply-line part of a rule is the part the agent skips; the marks that hold are steps (a grep, a version read, a function opened — #570's producer read). A test that asserts the quote is a step: it is a tool call before the edit, it is in the diff, and it is graded by the same run that grades the fix. "Never a looser test" closes the other way round the rule: a test written to the wider reading.

The loop also found the instance ungradeable in its environment (the gold itself grades 0 — Django 2.2 on SQLite ≥ 3.26), and knows no other instance where the request quotes its reference. So this cannot be measured until one turns up; the loop reads first when it does. The change is the shape the loop's evidence supports, at no new tokens.

## Test

```
test prompt::role_tests::a_quoted_reference_fixes_what_done_means_and_the_test_is_its_mark ... ok
test result: ok. 8 passed (prompt::)
```


unarbos:
Merged into main as d2a807e4 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is d2a807e4.

PR #632, opened 2026-09-18
loop: the naming check walks every screen, catches one-word symbols, and leaves the app on the list

unarbos:
Harness only. Nothing under `ios/` moves, so the iPhone build job skips.

`check-names.sh` (#626) only ever walked the projects list and a chat. This takes it to the rest of the app and closes two holes it had.

**The other three screens.** It now visits the workers sheet, the settings sheet and the call. Settings has more controls than any other screen and produced the first symbol name anybody noticed; the call shows almost no words by design and so rests entirely on its labels. Both come back clean, which is the labelling work of M-264/M-297/M-314 holding where it was done.

**The sheet grabber is not ours.** UIKit draws and names the sheet's drag handle. No app can relabel it, and a report listing something nobody can fix teaches the reader to skim it. Exempted, with the self-test requiring the exemption to hold.

**A symbol name is not always two words.** The multi-word rule walked straight past `Circle`, the away card's bullet, sitting one row from a name it did flag — in output I had already read and called clean. Shapes and objects nobody names a control after now count on their own.

**It left the app in the call.** Ending on the preview call sent the following steps tapping at a screen with no composer; they reported about the call while believing they were in a chat. It returns to the list now.

Run on the simulator, all five screens:

```
detector self-test: caught 4 of 4 known-bad names, and let 'Search' and
the system's 'Sheet Grabber' through

== the projects list ==   every control here is named
== a project's chat ==    every control here is named
== the workers sheet ==   every control here is named
== the settings sheet ==  every control here is named
== the call ==            every control here is named

VERDICT: no control reads as a symbol name
```

The chat is clean there because the build under test carries the `ios/` fix, which is batched separately under Apple's daily upload cap.

Findings M-358, M-359, M-361, M-362.


unarbos:
Merged into main as 4c52047d (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 4c52047d.

PR #633, opened 2026-09-18
mobile: the transcript ends above the workers pill, and two decorations stop speaking

unarbos:
The second `ios/` batch, held together because Apple refused the #615 upload with `Upload limit reached. Please wait 1 day`. One archive, three fixes.

## The transcript ended behind the pill

Scrolled to its end, the chat's last line sat **59 points** too low. The "While you were away" card — the one thing on screen asking to be read — arrived with `root replied / OK` under the workers pill.

The composer is a bottom inset so the transcript ends above it, but the pill row above the composer was not counted. Measured on the simulator:

| | last line, as opened | true end of content |
|---|---|---|
| chat with workers (pill) | y 731 — behind the pill | y 672 |
| chat with none (no pill) | y 721 | y 721 — identical |

The right-hand column is the diagnosis. With no pill, `scrollTo` lands exactly at the end. With one, it stops a pill's height short, and whatever is last sits behind it — the away card here, but equally a reply.

Not a timing fault. A settle-then-scroll changed nothing and a log line proved the scroll ran; a second scroll 400 ms later landed in the same place; a cold launch landed there too; a manual swipe went 59 points further every time. Two commits here are that wrong diagnosis and its removal, left in place rather than rewritten.

The fix reserves the height the pill actually measures, through a preference key rather than a constant, and zero when there are no workers. After it the card opens fully in view: `OK` at y 632 against the pill at 731.

[The away card fully visible above the workers pill](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Ftmp%2Fo109%2Faway-clear-of-pill.png)

## Two decorations that spoke

Both found by `check-names.sh`, and both a glyph repeating what the text beside it already says:

- the worker line's `arrow.turn.down.right`, read aloud as **"Arrow Turning Down Then Right"** before every worker in the chat;
- the away card's bullet, read aloud as **"Circle"**. Its colour marks an ask or a failure, and the words name which.

## Checks

- The naming check over all five screens: `VERDICT: no control reads as a symbol name`.
- Tail behaviour unchanged — a sent message's reply lands at y 319 with the pill at 437.
- A chat with no workers reserves nothing and is untouched.

Findings M-352, M-359, M-360.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

[...]

PR #638, opened 2026-09-18
mobile: a chat opens at its end, and loses the gap #633 left

unarbos:
This both fixes the real fault and takes #633 back out. #633 was mine, yesterday, and its stated mechanism was wrong.

## What is actually wrong

Opening a long chat lands **501 points** short of the end — about five lines of the newest reply, under the composer and the pill. Measured with `onScrollGeometryChange` rather than inferred from element positions:

```
off 5467.3  end 5968.7  short 501.3  insetB 105.3   ← as opened
off 5863.0  end 6002.7  short 139.7  insetB 139.3   ← after one more scroll
```

The rows between are not realised when the first scroll runs, so the proxy scrolls to an estimate. A second scroll, once they exist, arrives at the end. That is the whole fix: one `scrollTo` 150 ms after the first.

## Why #633's reservation goes

#633 blamed the pill row's height and reserved it at the end of the transcript. That cured the away card by luck — any upward shift would have — and left every chat opening with a pill-high band of nothing under its last line. Removed, along with its preference key and geometry probe.

## Checks

On the fixed build, both chats open **exactly** at their true end; three swipes move the last line by one point:

| | as opened | true end |
|---|---|---|
| `phone` — has workers, so has a pill | y 684 | y 684 |
| `const` — no workers, no pill | identical | identical |

The last line sits just above the pill, with no gap and nothing clipped:

[A chat opening with its last line just above the workers pill](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Ftmp%2Fo109%2Fopens-at-the-end.png)

The core chat path is unchanged: card 1.0 s, reply 2.0 s, `Worked` at 11.1 s, composer cleared.

Findings M-368, M-369. This closes M-360 and M-366.


unarbos:
Merged into main as 2be74d2c (merge commit); rust fast-forwarded to the same SHA. Apple's daily upload cap is already hit today, so the ios-testflight run for this SHA is expected to be refused; TestFlight stays 1997 unless the run prints `uploaded`.

PR #639, opened 2026-09-18
prompt: the producer rule's reply-line mark goes; the read call is the mark (SWE-bench cycle 27, 0 of 10)

unarbos:
## What

One clause of `CONTRACT`'s producer rule. Before:

> … open the sibling (read it; the call is the record that you looked) and put one line in your reply — "producer: <Class>.<attr> added the way <Sibling> has it" or, when the producer truly cannot have it, "fallback in <caller>: <why the producer cannot>".

After:

> … open the sibling (read it; the call is the record that you looked); a fallback is for the one case where the producer truly cannot have it, and then the reply says why it cannot.

The step stays; the formatted reply line goes. Pin updated (asserts the line is absent and the fallback clause present).

## Why

`internal/features-inbox/2026-09-18-swebench-producer-step-read.md` (cycle 27, ten rollouts): the read call before the first edit (#570) was taken in 9 of 10 and is what moved xarray-6938 from 0 of 3 at the producer to 5 of 5 solved. **The `producer:` reply line appeared in none of the ten.** The loop's standing finding holds for every reply-line mark I have written (`no change:`, `as quoted:` — dropped in #627 — and now `producer:`) and for none of the step marks. A line the agent never writes is tokens spent on every worker turn for a mark no reader gets.

The cycle's other finding — the predicate hedge on sympy-17318 — is already on `main` as #608 ("when you can name the wrong predicate, change the predicate"); this PR does not touch it.

## Test

```
test result: ok. 8 passed (prompt::)
```


unarbos:
Merged into main as dc633b7d (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is dc633b7d.

PR #646, opened 2026-09-18
transcripts: a half-written last line a dead kernel left is cut at start, before anything appends (qal-j37)

unarbos:
## The bug (QA, `qal-j37`, found 12:55 today)

`append_events` mends a write that failed part-way (`drop_partial_line`), but only in the process whose own write failed. A kernel killed mid-write — SIGKILL, a power cut, the OOM killer — leaves the head of a line with no newline and nobody to cut it. `load_transcript` skips unparseable lines silently, so that line was invisible; and the **next kernel's first append ran onto it** and was skipped too. One event per crash, and which one depends on what the first batch begins with: a `wake` (which #450's two-writer detector counts to decide whether a place was double-served) or the person's own `user` line.

QA staged it: three whole events plus a half line, one turn → 6 readable, 1 unparseable, the new `wake` inside the unparseable one.

## The fix

- `arbos_core::files::drop_headless_tail(path)`: if the file's last byte is not `\n`, cut back to the last whole line, scanning backwards in 64 KiB chunks (a half-written tool output can be long). Returns the bytes dropped; `None` for a whole or absent file.
- The kernel runs it over every live agent's transcript **right after taking the place lock and before `bootstrap`** — bootstrap's own notice is an append — so no write can be in flight, which is the one moment the cut is safe with concurrent in-process appenders. Archived agents are not touched.
- It is said: `transcript_repaired` in `kernel.log`, and a notice on that transcript — *The kernel that wrote this record before ended in the middle of a line (N bytes, the head of one event). That half line was dropped so everything from here on reads whole; the event it began was lost with that kernel, not now.*

The existing in-process repair in `append_events`' error arm stays as it was.

## Tests

- `files::headless_tail_tests` — cut to the last whole line; a whole record untouched; a record that is one headless line becomes empty; a 200 KiB half line past the scan chunk is found; an absent file is fine.
- `headless_tail_e2e` — the bug's own shape: three events and a half line, the kernel starts (no unparseable line remains, the notice and log line are there), a turn runs, and **no event appended after the crash sits inside an unparseable line**: the `wake` and `user` lines both read, the three whole events stay, the half event is gone. And a whole record is not touched (no cut, no notice, no log line).

```
test files::headless_tail_tests::a_headless_last_line_is_cut_and_a_whole_record_is_left ... ok
test a_half_written_last_line_is_cut_at_start_and_the_next_turn_reads_whole ... ok
test a_whole_record_is_not_touched ... ok
```

`restart_states_e2e` and `tail_cursors_e2e` (the neighbours that restart kernels over existing records) still pass.

## Not done here

QA's two further injections — `ENOSPC` on the supported path, and a shifted clock via `ARBOS_NOW` — are the loop's probes to add; nothing in the kernel changes for them.


unarbos:
Merged into main as cecd48e1 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is cecd48e1.

PR #648, opened 2026-09-18
loop: the refusal check reaches the list again, and says what each case should do

unarbos:
Harness only.

## It had measured nothing since cycle 100

`refusal-and-transport.sh` ran, printed four sections, and ended with its paragraph about how a refusal and a transport failure should differ. Every section said `no <row> row`. It had opened no case at all.

The cause is mine. At cycle 100 I made a cold start come back to the chat that was in front (M-338). That was the right change, and it means the app no longer lands on the projects list — which is where this scenario taps its four fixture rows by name. Fifteen cycles of runs that looked complete and measured nothing, because the closing text prints unconditionally.

Three changes: it reaches the list, it refuses to run when the fixture's four rows are not on screen, and a case that never opens produces `VERDICT: none — N of 4 cases never opened, so this run says nothing`.

## The behaviour is fine

Once it could see the fixture again:

| case | attaches in 25 s | |
|---|---|---|
| `no machine named` | **1** | stops — nothing will change |
| `is offline` | 3 | retries — comes back when a kernel starts |
| `has no kernel serving` | 3 | retries — comes back when someone starts it |
| 502 at the tunnel | 3 | retries — nobody answered |

Identical to M-257 at cycle 70. Only the instrument had rotted.

## The closing paragraph contradicted its own finding

It read "a refusal should be counted once and left alone; the 502 should be counted several times" — so by its own words two of the three refusals were faults. M-257, which this file was written to record, measured 1, 3, 3, 3 and called the rule held.

Two refusals *should* retry: they name a thing that comes back, an offline machine or a kernel not started. Only an unknown machine is final, and the app's own wording says so — "fixture-box has been off for 11 hours. It comes back when a kernel starts on it. Retrying in 11s."

Each case now carries its expectation and the run reaches a verdict, so nobody judges correct behaviour against a cruder rule than the app's:

```
VERDICT: final where nothing can change, patient where it can —
         an unknown machine is asked once; an offline machine, a
         missing kernel and a dead tunnel are all waited out
```

Findings M-379, M-380, M-381.


unarbos:
Merged into main as 5de5f3f5 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 5de5f3f5.

PR #651, opened 2026-09-18
loop: every scenario that taps a project row reaches the list first

unarbos:
Harness only. Follows [#648](https://github.com/unarbos/arbos/pull/648), which found one scenario with this fault; this is the sweep for the rest.

## Seven more had the same assumption

`chip-and-send-arrow`, `photo-reaches-the-model`, `pill-count-vs-sheet`, `several-workers`, `the-core-chat-path`, `voice-notes-wait` and `worker-while-it-works` all launch the app and tap a project row by name. Since M-338 a cold start comes back to the chat that was in front, so there may be no list to tap.

**They were not all broken, which is worse.** A fresh install has no front project and lands on the list. So a scenario passes when it runs first and taps at a chat when it runs after another — the answer depends on what ran before it, and nothing says so. `cold-start-and-history` still reports `3.0s to a list of 7 rows` for exactly this reason: it runs straight after the install.

One `reach_the_list` in `sim-lib.sh`, not seven copies, because eight copies of an idiom is how the next one gets missed.

Proven by running `the-core-chat-path` twice back to back, so the second run starts with a chat restored. Both reach the same verdict now.

## It passed while blind on its first test

Run from a shell where `BASH_SOURCE` is unset, the helper resolved its own directory to `.`, could not find `ui.py`, got an empty dump, read "no Back button" from that and returned success — with the app in a chat the whole time, as the next dump showed.

That is the same fault as the one it was written to fix. An empty tree is "I cannot see", not "nothing is there", and it says so now:

```
--- woke up in:   Back buttons: 1
reach_the_list: ok
--- after:        Back buttons: 0, project rows: 12
--- blind:        guarded: refused to answer blind
                  cannot read the screen — is /nowhere/ui.py there, and the app up?
```

Findings M-382, M-383.


unarbos:
Merged into main as 1beec0a1 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 1beec0a1.

PR #662, opened 2026-09-18
loop: check that the step happens, not that the file mentions it

unarbos:
Harness only. Corrects my own sweep in [#651](https://github.com/unarbos/arbos/pull/651).

## The sweep used a proxy

Cycle 116 looked for scenarios that assume the app opens on the projects list by searching each file for the string `Button +Back`, and marked four of them clean.

`call-pulled-down` matched that string **inside an unrelated helper** — a `where()` that builds a grep from it — and has been printing `no phone row` in every sweep since. `notifications`, `worker-chat-open-and-back` and `what-a-returning-user-sees` were cleared the same way; they happened to keep working because the chat they come back to is usually the row they want.

A file mentioning a thing is not a file doing it. The check now lives in `check-tools.sh`, beside the other harness-hygiene checks, and asks whether the step happens *before* the tap, in line order.

## It found one more the moment it ran

`cold-start-and-history` taps its row with no step to the list. Worse, its first measurement was of a behaviour the app no longer has — "launch to a list with rows" — which only worked because the sweep runs it first, on a fresh install with no chat to come back to.

It waits for either landing now and names which:

```
VERDICT cold start:   3.3s to the chat it was left in (pod)
                      (cycle 37 measured 3.1s to a list)
```

The number stays comparable to cycle 37's; the sentence stops being false.

## Eleven scenarios, eleven conclusions

With #651 on `main` and these fixed, the sweep is whole: `call-pulled-down` concludes again with `pulled down it types, mutes, and closes back to the chat it came from`, and `voice-notes-wait` and `pill-count-vs-sheet` came back on their own.

One flap worth watching, not fixed here: `list-search-filter-refresh` said the refresh worked in one run and `the list is not redrawing the answer` in the next. Its own header warns that its fixture's trigger is timing-bound.

Findings M-389, M-390, M-391.


unarbos:
Merged into main as edf8dc77 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is edf8dc77.

PR #665, opened 2026-09-18
loop: log what the fixture answered, not only that it was asked

unarbos:
Harness only. Closes the flap flagged in [#662](https://github.com/unarbos/arbos/pull/662).

Two runs of `list-search-filter-refresh` an hour apart disagreed about whether pull-to-refresh works, and neither could say why. The check counted `/list` calls before and after the pull, and the count rose both times.

A rise is consistent with two quite different stories:

- the app asked again **before** the fixture began serving the new project — the request was already in flight, so the run tests nothing;
- it was answered with the new project and did not draw it — a real fault in the list.

The scenario blamed the app either way. The fixture now records whether each answer carried the project:

```
/list calls:          2 before the pull, 3 after
answers carrying it:  1 of 1 calls since the pull
  VERDICT: the pull refetched and the new project reached the screen
```

and a run where none did says so instead of naming a fault:

```
  VERDICT: cannot say — the app asked again, but every answer since the pull
           was sent before the fixture began serving the new project. The
           request was already in flight; this run tests nothing.
```

**The flap did not reproduce.** Four runs — three isolated and one after `list-composer`, as the sweep orders them — all pass with the answer carrying the project. I stopped rather than hunt a fifth green: the next occurrence will name which story it was, which is worth more than another pass today.

Findings M-392, M-393.


unarbos:
Merged into main as c96951db (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is c96951db.

PR #670, opened 2026-09-18
loop: a decline that could never stop declining, and the accusation it made when it could

unarbos:
Harness only.

## The decline was permanent

`pill-count-vs-sheet` ends every sweep with `cannot say`. Its reason is correct — while anything runs the pill reads `Working N`, counting only what runs, and the sheet lists every agent, so the two are a subset against a whole.

But in the sweep it runs after the scenarios that start workers, so something is always working. A row that declines every time is uncovered while looking careful. It now waits up to three minutes for the pill to read `Agents N`, and says how long it waited if it declines anyway.

## Then it accused the app

The first run past the wait said:

```
VERDICT: they disagree — pill 38, sheet 30, and no shared labels to explain it.
```

The app cannot disagree with itself here. The pill is `chat.workers.count`, the sheet is a `ForEach` over that same array, and `workers` is a dictionary keyed by the agent's id, so no row is lost to a collision. Reading the source settled in a minute what no amount of re-running would have.

## Which rig fault, told apart by the page count

`collect_rows` now reports how many pages converging took, because "converged after 1 page" and "converged after 9" are different stories about the same word. Here: `converged after 5 page(s)` — the sheet scrolled and still stopped eight short, so the swipe is fine.

Rows are counted by their **label**, and this project has eight more agents than distinct goals. Twins on different pages collapse under `sort -u`, and the duplicate check only ever looked within one screen.

Confirmed on a project whose goals are distinct:

```
qa-cycle-11-demo — the pill says:  Agents 12
                   the sheet lists: 12 rows (paging converged after 2 page(s))
                   VERDICT: the two agree on 12.
```

The verdict now separates "never scrolled, fix the swipe" from "scrolled, and the labels repeat".

Findings M-397, M-398, M-399.


unarbos:
Merged into main as 75a00502 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 75a00502.

PR #642, opened 2026-09-18
mobile: a scroll by hand cannot rest under the pill, and the list stops speaking its separator

unarbos:
The third `ios/` batch. Two fixes, one archive.

## A scroll by hand came to rest under the pill

#638 made a chat *open* at its end. Scrolling there **by hand** was still wrong: the reader's own scroll came to rest with the last line under the workers pill, and they had to nudge it out.

I only found this by watching a recording and asking what was literally on screen:

> scrolling back down rests with the last message **partially hidden** behind the pill, and the user has to perform an additional upward swipe to pull the rest of the message into view.

Four stills across three cycles had missed it, and they always would: a still shows where a scroll *landed*, never where a scroll *rests*.

A bottom content margin the height of the pill fixes the resting place. The opening position is untouched — the last line sits at y 581 both as opened and at the true end.

[A chat after scrolling to the bottom by hand, last line fully clear of the pill](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Ftmp%2Fo109%2Fmanual-rest.png)

## The list spoke its separator

The `phone` row read `phone, Idle,  · , home, 17m`. The " · " between the state and the machine is a `Text` of its own, so SwiftUI hands it to VoiceOver as an element. Only `phone` showed it, because only `phone` has a machine part.

Caught by `list-rows.sh` ([#641](https://github.com/unarbos/arbos/pull/641)) on its first run, and clean after this.

Findings M-370, M-372.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

[...]

PR #673, opened 2026-09-18
loop: the list never emptied — the check was counting rows by their status

unarbos:
Harness only.

`settings-and-a-bad-token` has ended every sweep with:

```
VERDICT: a token that cannot work empties the list to 1, says why, and 7 come back
```

It does not empty the list, and never did. A review of the bad-token recording said so flatly:

> the Projects list never shrinks or collapses. There are still exactly 7 rows. Instead of disappearing, the status text under the project names changes to "Off".

## The cause

`rows()` counted only rows matching `<name>, (Idle|Working)`. A refused token leaves every row where it is and changes what each one says, so the six rows that turned `Off` stopped matching the pattern and were counted as gone. A sentence about the app that was a fact about a regex.

## Measured on today's build

```
  rows now: 7 (1 still reading Idle or Working)
  what the rows say: 6×Off  1×Idle
  what the screen says:
      Hub token refused.
  rows after restoring the token: 7

VERDICT: a token that cannot work keeps all 7 rows and marks them off
         (1 still live), says why, and 7 come back
```

The one still live is the local kernel, which does not go through the hub.

The scenario counts project rows regardless of status now, reports how many still read live and what the statuses became, and keeps a separate branch for rows genuinely disappearing — which would be a change from this behaviour, not a confirmation of it.

## Worth noting

This is the second time this week a recording has beaten the accessibility tree ([#642](https://github.com/unarbos/arbos/pull/642) was the scroll coming to rest under the pill). Both times the tree was read correctly and the *question asked of it* was wrong: "how many rows say Idle" is not "how many rows are there", and a still cannot show where a scroll rests. Neither check would ever have failed on its own.

Findings M-400, M-401.


unarbos:
Merged into main as 97c1b89b (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 97c1b89b.

PR #674, opened 2026-09-18
loop: the cold start counts rows, not rows that read Idle

unarbos:
Harness only. Follows [#673](https://github.com/unarbos/arbos/pull/673), which found one count filtered by status and described as a count of rows. That shape can be grepped for, so this is the audit.

Three counts across the scenarios carry a status word in their pattern. **One was the same fault.**

`cold-start-and-history` waited for a row reading `Idle` or `Working`, counted those, and called the result "a list of N rows". On the accessibility tree this loop captured at cycle 112:

| | |
|---|---|
| rows, any status | **12** |
| rows reading Idle or Working | **7** |
| what the other five said | `mac is asleep` |

So "a list of 7 rows" for a list of twelve. And a list whose machines were all asleep would have waited the full 40 seconds and reported `never` — a timeout presented as a cold-start measurement.

It counts project rows now and names how many are live beside the total:

```
launch to a list of 7 rows (7 live): 3.1s   (cycle 37 measured 3.1s to a list)
```

**The other two were already honest**, and that is worth recording too. `worker-while-it-works` reports `sheet rows after scrolling to the end: N   of them live: M`, naming both quantities. `settings-and-a-bad-token`'s `live_rows` is deliberately status-based and is called that.

The difference is not care. It is that they say which quantity they are reporting — which is the whole lesson of the last four cycles in one line.

Findings M-402, M-403.


unarbos:
Merged into main as e1362844 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is e1362844.

PR #677, opened 2026-09-18
loop: barge-in is a committed check, and it reads the reason the app prints

unarbos:
Harness only.

## The row was a memory

Barge-in was measured at cycle 77 with ad-hoc commands and never written down. That is the trap M-270 named when the mic rig had to be rebuilt from scratch for the same reason, and it has stood for 47 cycles.

`barge-in.sh` reads the two metrics the app already emits — `barge_in_speech_started` and `barge_in_response_done` — rather than timing anything itself:

```
run 2: it stopped talking after 192 ms, the gateway confirmed at 237 ms
run 4: it stopped talking after 335 ms, the gateway confirmed at 381 ms

runs that could barge in at all:  2 of 4
runs that stopped the reply:      2 of 2
VERDICT: speaking over it stops it, every run that could — 263 ms on average
         (cycle 77 measured 181 ms)
```

## Its first version filed a fault that was not there

It reported `1 of 3 — not every time, which is worse than never`. The app had written the reason for both other runs and I had not read them:

- `barge_in_skipped reply already over` — the reply finished before the clip fired, so there was nothing to interrupt;
- `barge_in_not_armed clip already used` — the clip was spent on an earlier reply.

Neither is barge-in failing. Runs that never had a chance are counted apart from runs that did, so the verdict is about the runs that could test the thing.

## Two things found along the way, not changed here

**The orb does change with the phase.** A review of the recording reported it "completely static" for 30 seconds. Measured off the stills, the orb region reads `(127,127,126)` listening, `(94,94,93)` thinking, `(71,88,111)` speaking — dimmer and blue. The review was right about that recording and wrong about the orb: that call spent most of its length in one phase.

**After a barge-in the call says "thinking" about a reply that never comes.** Sampling the phase every two seconds against the console: barge at 202 ms, gateway confirms at 247, `response.done reason=completed playing=false`, then `thinking` for the remaining 14 seconds with no further events, until the caller speaks again. `settle()` runs only when a reply reached the caller, and the code says why — "a reply that reached nobody must not take the screen back to listening". That is true, and it swaps one wrong statement for another. What it should say instead is a design call, so it is recorded rather than changed.

Findings M-404 through M-407.


unarbos:
Merged into main as 00cc5ba8 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 00cc5ba8.

PR #680, opened 2026-09-18
loop: check the precondition this one only stated — and correct the reason I gave for it

unarbos:
Harness only. Two commits, and the second corrects the first on purpose.

## The row holds

`the harness — typing into the composer` was last measured at cycle 72. On `main`: **4 of 4 typed lines arrived, 4 of 4 whole**, against M-162's six-of-eight loss in the cycle-48 journey. It ends with a verdict line now, so anything reading it in bulk can, rather than leaving a reader to parse `new: 4/4 arrived, 4/4 whole`.

## A precondition stated in a header and never checked

"Run it with the app on a project chat" was a comment and nothing more. The screens with no composer — the call, settings — would have produced a count of nothing typed. It refuses without a composer now.

## And then I invented the failure it prevents

The guard's first comment explained what would otherwise happen:

> Run from the projects list there is no composer to focus, every line goes nowhere, and the run reports `0/4 arrived`.

Then I ran it from the projects list. **2 of 2 arrived, 2 of 2 whole.** The list has its own composer, `Message phone…`, and it sends to the project it names.

The guard was right; its reason was invented — inside a change whose entire point is that a precondition should be checked rather than asserted. The second commit corrects it rather than amending, because the wrong sentence is the useful part of the record.

That also measures something that had never been measured: typed lines sent from the **projects list** composer reach the kernel whole.

Findings M-408, M-409, M-410.


unarbos:
Merged into main as 4db6f8ee (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 4db6f8ee.

PR #683, opened 2026-09-18
loop: the journey needed the list too, and the check that guards it did not work

unarbos:
Harness only.

## The row holds

A full journey run against `pod` on kernel `c3247332`: **P1, P2, P2e and P3 all pass**, alongside J1–J3, J4a/J4m, J5q/J5s and J7/J7v. The photo step passes on the model naming the picture — "Ice plant flowers, predominantly magenta" — not on the reply failing to sound like a refusal, which is how it passed for months while attaching nothing.

## But J1 did not open a project

"Open the project from the list" tapped `phone` at **y=85** — the chat's own header, not a row — and passed, because the chat the app woke in happened to be the target. The journey has the fault the last several PRs have been clearing out of the scenarios, and the check written for it never looked at the journey, which lives outside `scenarios/`.

## And the check stayed silent with the step deleted

Pointing it at the journey was not enough. It has been wrong in both directions at once:

- **A one-line helper.** `score() { …; }` opens a brace and closes it on the same line. The body-skipping took that as entering a function and ignored every use in the rest of the file.
- **The last reach, not the first.** It recorded the final `reach_the_list`, so a file that reaches the list at J1 and again after the J6k relaunch looked like one that never did.

Both fixed, and then demonstrated the only way that means anything:

```
--- as committed:
  all of them reach the list first
--- with the journey's first step deleted:
  mac-journey.sh    needs the list at line 109, with no reach_the_list before it
```

This rule has been narrowed and widened across three cycles now. The version I trust is the one I watched fail on purpose.

`sleeping-machine` was the last scenario genuinely missing the step.

Findings M-411, M-412, M-413.


unarbos:
Merged into main as 107ed134 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 107ed134.

PR #685, opened 2026-09-18
loop: say when the argument is the wrong kind, not that there is nothing

unarbos:
Harness only.

## The row holds, with 31 more connects

`call — the microphone path` was last measured at cycle 79, where the long-running "connect-time drift" look was settled: never drift, two engines. Re-reading every call log the loop has written:

| engine | n | min | median | p90 | max | mean |
|---|---|---|---|---|---|---|
| duplex | 101 | 284 | 543 | 732 | 898 | 526 |
| openai | **68** | 999 | 1951 | 2800 | **5773** | 2089 |

Still no overlap — the slowest `duplex` connect is faster than the fastest `openai` one. `openai` has 31 more samples than at cycle 79 and `duplex` none, which is the phone having moved to GPT Live.

**The useful number is the max, and it has not moved.** 5773 ms then, 5773 ms now, across 31 further connects. So the 6000 ms hold's 227 ms margin is thin but *stable* rather than eroding — a different thing to worry about, and worth knowing before anyone spends a cycle widening the hold.

## The change here

This scenario takes an output directory where every other one takes a cycle number. Given `127` it answered:

```
no logs under 127
```

which reads as "there are no logs", not "that is not what I take". I fell into it on the first run of this cycle, and a reader skimming a sweep would have taken it for an empty record.

A bare number that is not a directory is now named for what it is:

```
'127' looks like a cycle number. This one takes an output root,
not a cycle: it reads every call log the loop has ever written.
Run it with no argument to read them all.
```

Findings M-414, M-415.


unarbos:
Merged into main as f47d7ed3 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is f47d7ed3.

PR #631, opened 2026-09-18
desktop: the kernel's reason replaces the window's 'no reply from the kernel' guess (F-189)

unarbos:
## What

Cycle 41, after-failure states on the new kernel: the API key goes bad in `config.toml` while the kernel runs. The kernel re-reads the key per turn (good) and refuses the turn with *did not accept the API key*. The window drew **two** lines under the prompt: *no reply from the kernel*, then the kernel's reason.

## Why

*no reply from the kernel* is the window's line for a turn that ended with nothing under the prompt (`TurnDone(Ok)` with the prompt as the last item). The kernel's failed notice lands a tail-tick later as `TurnDone(Err)`, so both stood.

## Change

`notice()`: a failed notice arriving right after the guess takes its place. The guess stays when nothing follows it.

## Measured

`keygone.py` (fresh place, kickoff, then a dead key written to `config.toml`, then a prompt). Before: `[user, notice "no reply from the kernel", notice "turn failed: … did not accept the API key"]`; after: `[user, notice "turn failed: … did not accept the API key"]` — still in `media/cursor-reference/cycle-41/key-gone/01-fixed-crop.png`. Removing the key line altogether falls back to the kernel's environment and the turn answers; nothing to pair there.

Gate `cycle-41d` (`ui_pass.py`, all phases, kernel `883a47da`): **135 pass, 1 fail** (`composer-context`, old) — the first clean full gate on the rebuilt box.

No `v0.2.0` publish.

<div><a href="https://cursor.com/agents/bc-2a1318aa-e675-52f4-b3ab-94cb9415aa39?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2a1318aa-e675-52f4-b3ab-94cb9415aa39&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #689, opened 2026-09-18
loop: what a worker's chat is made of, counted rather than looked at

unarbos:
Harness only. This was the last coverage row still judged by eye.

## The claims were always about the tree

`style pair vs Cursor stills — worker chat` has been read by eye since cycle 37: "plain header, no composer, the header reads Back". Every one of those is a statement about the accessibility tree.

```
opened: count slowly one to forty, Done

  the header names                   count slowly one to forty
  back control reads                 1
  composers (M-154: none)            0
  send buttons                       0
  microphone buttons                 0
  controls in the header             1
  text rows on screen                5

VERDICT: the shape holds — one Back, a name, no composer, and nothing
         else in the header
```

`style-pair.py` cannot do this row: it measures row pitch and ground colour, which are properties of a **list**, and it refuses this pairing since [#631]. That refusal was honest and left the row unmeasured. This is the part that can be measured.

## It fails on the neighbouring screen rather than passing

A shape check that would pass on the project's chat measures nothing, so I ran the same counts there. One Back away: **1 composer and 3 header controls**, against the worker chat's 0 and 1.

The absent composer is a deliberate choice — a dead input is worse than none — so a `TextField` appearing here is a regression, not an improvement, and the check treats it as one. It also proves it is in a pushed chat before counting anything about one.

## A smaller thing

`pod` left the roster between two runs an hour apart, and the scenario said `no pod row` — which reads as the taps having failed rather than a project having gone. It prints the list it did see now, and defaults to `phone`, the same kernel drawn twice and the row that is actually there.

Findings M-416, M-417, M-418.


unarbos:
Merged into main as 2f517129 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 2f517129.

PR #690, opened 2026-09-18
loop — the Jev hop draws nothing, and a folder question is one listing and then an answer

unarbos:
## What

Two fixes to the Jev hop in `crates/arbos-engine`, both from one reported turn on the Mac (build 2083, project Clawdbot, Jev on, `what files are in this folder?`):

1. Asking Jev no longer writes a status line. `Choosing the next step` is gone from the chat face. Jev still runs and still picks the next move.
2. A folder question runs `ls` once and then answers. It was running `ls` as `jev-1` and again as `jev-2` through `jev-415`, with `grep` and `read` picked in between.

## Why

**The face.** `jev::ask` set the derived status line to `Choosing the next step` on every ordinary turn, so the window narrated the kernel's own plumbing instead of the work. Jacob rejected that line. Nothing replaces it: the chat shows the tool Jev picked and the answer, exactly as it does for a step the model chose. No waiting banner either — the hop passes a no-op where it used to pass `hooks.working`. One heartbeat is 5s and Jev's first-byte cap is 1.5s, so that callback never actually fired; the test now pins `FIRST_BYTE < HEARTBEAT` instead of a bare 3s, so widening the cap cannot bring a banner back.

**The storm.** All three rules below come from one fact: the Decisions door answers with a tool name and no arguments, so `Decision::args` is always `{}`.

- **Jev is only offered tools a pick can run.** A tool with a required argument cannot be run from a pick that carries none. `grep`, `read`, and `find` are no longer choices, so the card cannot produce a search storm; `ls` is a choice, and its default path is the open folder. Searching is now a language step, and Jev picks `llm` for it. The situation card's `tools:` line and the Decisions `tool` criteria both come from the same `jev::pickable(&view)` list.
- **A pick missing its required arguments routes to the model** instead of running. Running it only recorded "missing argument" and left the same card in place for the next hop to pick from again.
- **A pick the turn already ran with the same arguments routes to the model.** The listing is already on the transcript; what is missing is the sentence that reads it back. This is the cap: `jev::Picks` lives for one turn and remembers each `name args` it let through.

`undo` needs no argument and restores the checkpoint, so it is never a pick. It was one choice among thirty before this change and would otherwise now be one of four. The model still asks for it by name.

Evidence for the reported turn: `/cursor/stores/bc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983/internal/jev-mac-live-check.md`.

## How to check

- [x] `cargo test --workspace` passes — 160 test binaries, exit 0, no failures.
- [ ] `cd desktop && cargo build` — not run: `yeslogic-fontconfig-sys` cannot find `fontconfig.pc` in this container and the package is not installable here. The desktop crate is unchanged and depends on `arbos-core` and `arbos-update` only, not on `arbos-engine`, so the changed API is out of its reach.
- [x] No manual check on the Mac from here. The turn's behaviour is pinned by tests in `crates/arbos-engine/src/jev.rs` instead:
  - `a_pick_with_no_arguments_only_offers_tools_that_need_none` — the offered list holds `ls` and not `grep`, `read`, `find`, `write`, `apply_patch`, `say`, `ask`, `plan`, or `undo`.
  - `a_tool_missing_its_required_arguments_becomes_llm` — a real Decisions `answers` body picking `grep` parses to `args == {}` and routes to `llm`; the same body picking `ls` routes to `Tool { ls, {} }`.
  - `the_same_call_twice_is_one_listing_and_then_an_answer` — the first `ls {}` runs, every repeat routes to `llm`, and `ls {"path":"src"}` is treated as a different call.
  - `a_pick_of_undo_becomes_llm`.
  - `first_byte_is_a_router_wait_not_a_chat_wait` — updated to pin the cap against `HEARTBEAT`.

## Notes for the reviewer

- The repeat rule is scoped to Jev's picks. Tool calls the language model writes still go through `model_step` and `batch`, untouched, so a model re-running a test command in one turn is unaffected.
- `Picks` compares the rendered `name args` string. That is exact-match only, which is all the Decisions door can produce today. If Jev ever gains the ability to send arguments, this guard gets looser on its own rather than tighter — worth revisiting then, not now.
- `NEVER_PICKED` is a one-name list. A read-only test derived from each tool's `Plan`/`Access` would be more principled, but `plan` needs a `PlanCx` that `pickable` does not have.
- Out of scope, as asked: no v0.2.0 publish, no Jev slices A–G, no qal-j35.

<div><a href="https://cursor.com/agents/bc-c1b06b11-d50d-5287-93cd-bffca4a7becc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c1b06b11-d50d-5287-93cd-bffca4a7becc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merged into main as c873ca04 (merge commit) on the coordinator's word, CI green on 27cf3ef9; rust fast-forwarded to the same SHA. Mac update channel carries it: new main is c873ca04.

PR #692, opened 2026-09-18
Terminal and Browser open promptly

unarbos:
## What

Opening a Terminal and opening a Browser from the window. Two separate causes, one on each side of the socket.

**Terminal** — the pane read a connection it opened itself, and that connection was always too late for the prompt. It now reads the one the window already holds.

**Browser** — the kernel ran Chromium's cold start on its one frame loop, and the blocking HTTP client on that path panicked it. The click killed the place's kernel. That work is on a blocking task now.

### Terminal: the pane's own socket could never hear the prompt

A pane called `attach_or_spawn_place` on every open — canonicalize the folder, bootstrap `.arbos/`, probe the attach port, `GET /healthz`, compare the running kernel's commit with the bundled one — then opened a second attach, which the kernel answers with a greeting, a snapshot, a plan frame per agent and the focused chat's transcript tail.

After all of that the prompt was not there to be had. The kernel broadcasts a shell's output to every client attached to the place; the prompt had gone out over the window's own connection, moments before the pane's socket existed. The pane's `pty_in` carries no bytes, so nothing made the shell print again. What the person saw was "Connecting…", then "Connected" over a black rectangle, until they pressed a key.

The output is kept per page on the project now (`desktop/src/model/pty.rs`) as it arrives on the window's connection, and handed to the pane when it mounts — whole, prompt included — and after that as it comes. Keys go back the same way. The pane opens nothing and waits on nothing.

Two details worth a reviewer's eye:

- Every chat in a place holds its own attach socket and the kernel broadcasts to all of them, so the same output arrives once per chat. The first chat to deliver a page claims it and the rest are dropped; the claim is released when that chat's socket goes.
- Output is a `PtyWrote` event, not a `notify`. A shell printing a build log writes many times a second, and notifying the model repaints the whole window — transcript, sidebar and all — for each of them. The window hears the event and feeds the one pane.

### Browser: the click killed the kernel

`Frame::Browse` ran the whole of `BrowserHub::act("navigate")` inline on the kernel's frame loop: spawn Chromium, poll for its devtools port, open a WebSocket, wait for a page load, read the title, run the page-snapshot script. Every client frame in the place — a keystroke into a terminal, a message to an agent — queued behind it.

Worse, it did not finish. `reqwest::blocking` drops a Tokio runtime, and dropping one on a runtime thread panics:

```
thread 'main' panicked at tokio-1.53.1/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed.
```

So the first Browser click took the place's kernel down with it, and the window then had to notice and start a replacement.

Three changes on that path:

- the browse runs on a blocking task, so the kernel goes on serving;
- `ensure_chrome` serialises its start. A window's `browse` and the screencast it turns on arrive a frame apart, and both used to launch: the second deleted the `DevToolsActivePort` the first had just written, so neither found a port and both waited out the 25-second deadline;
- the window's open goes to the page and takes one picture on one CDP connection. `act("navigate")` also reads the title and runs the snapshot script — two more round trips and four kilobytes of page text, gathered for a model that is not being spoken to here.

## How to check

- [x] `cargo test --workspace` passes
- [x] `cd desktop && cargo test` passes (101 lib tests, 1 integration test)
- [x] CI green on `5b65c1cc`, all six checks
- [x] Measured before and after with the same two harnesses, on the same machine, against real kernels

### The harnesses

`crates/arbos-kernel/tests/open_speed_e2e.rs` drives a real `arbos-kernel serve` over the attach socket and times the frames a window waits for. `desktop/tests/open_speed.rs` times both terminal routes — the window's own connection and a socket of the pane's own — against one live kernel, off one clock.

Both were run at `4db6f8ee` (the base) and at this branch, with the harness files identical in each. `--nocapture`; every line starts with `MEASURE`.

### Browser

| measure | before | after |
| --- | --- | --- |
| `browse` → `board` (the tab can be drawn) | 41 ms | 47 ms |
| a `shell` asked one frame behind the `browse` | **never — the kernel had panicked and exited** | 47 ms |
| `browse` → first picture of the page | **never — same reason** | 0.3–1.2 s |

The middle row is the one that says the kernel is still serving while Chromium comes up. The last is Chromium's own cold start, which is not ours to remove; it swings with how warm the machine is on Chromium's files (341 ms warm, 1182 ms on the first start after a boot).

### Terminal

| measure | before | after |
| --- | --- | --- |
| `shell` → `board` | 44 ms | 43 ms |
| `shell` → prompt on the window's own connection | 207 ms | 200 ms |
| `shell` → prompt on a socket of the pane's own | **none in 5 s** | **none in 5 s** |
| `attach_or_spawn_place` preflight, warm | 6 ms | 8 ms |

The kernel was never the slow part here, and the two routes behave the same on both commits — which is the point of measuring them side by side. What changed is which route the pane is on. The 200 ms is the login shell starting and printing; the pane now has those bytes as soon as the window does, instead of an empty screen and a wait for a keystroke.

The preflight is small on this machine — an empty scratch place, a warm page cache, Linux. On a real project it is a canonicalize, a `.arbos/` bootstrap including a git exclude, a TCP probe and an HTTP round trip, and it ran on every terminal open. It is not on that path any more.

## Notes for the reviewer

- **The first CI run was red, and it was the harness's fault.** `port_health_e2e`, which times a kernel's startup, failed with a broken pipe. The browse measures started two kernels, hence two Chromiums, and a test that ends with `child.kill()` sends SIGKILL — so the kernel never returns from `run`, never drops its `BrowserHub`, and the browser it started outlives it. Seventeen Chromium processes stayed resident for the remaining twenty minutes of the suite, on a runner with four cores. `5b65c1cc` answers all three browser questions from one kernel and one Chromium, and stops its kernels with a signal they handle, which takes the browser and its profile directory with them. Measured after that change: no chrome process and no `/tmp/arbos-chrome-*` left behind. The suite is green.
- The same SIGKILL leak is in the older browser tests (`browser_e2e`, `browser_takeover_e2e`); I left them alone rather than widen this PR, but it is worth a follow-up.
- I could not run the Mac app here, so the terminal change is covered by the desktop test suite and by reading rather than by a click. The riskiest part is `Arbos::sync_terminal` and the new `feed_terminal` in `desktop/src/view/root.rs`.
- Chromium's cold start is still there on the first Browser open of a kernel's life. Pre-warming it when the drawer's tab cards are shown would hide that, but launching a browser because somebody opened a tab picker is speculative work; I left it.
- The terminal's scrollback is capped at 128 KB per shell and dropped when the row closes.
- Not touched, per the brief: v0.2.0 is not published, Jev, the chrome clutter list, and qal-j35.

<div><a href="https://cursor.com/agents/bc-fddc4e9f-b8ac-5d93-a0fa-0cb3ec84a180?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fddc4e9f-b8ac-5d93-a0fa-0cb3ec84a180&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


unarbos:
Merged into main as 4c7b938c (merge commit) on the green head 5b65c1cc; rust fast-forwarded to the same SHA. Desktop cargo check passed on the merged tree. Mac update channel carries it: new main is 4c7b938c.

PR #693, opened 2026-09-18
mobile: the chat does not say "Working Thinking"

unarbos:
The sixth `ios/` batch. One wording fix, found by watching rather than counting.

## What the recording caught

A recording of the photo flow — reviewed for what is literally on screen — showed the live line under the sent photo reading:

> Working Thinking · 5s

The chat prefixes `Working` to whatever step the kernel sends. `Starting` and `waiting on …` had each been special-cased when they turned up. `Thinking · 5s` had not.

## The rule the exceptions were hiding

The kernel sends two kinds of step and they need opposite treatment:

- a **state it has named** — `Starting`, `Thinking · 5s` — is already a sentence and should be shown as it is;
- an **activity** — `bash · cargo test`, `waiting on demo` — is a thing being done and takes a word in front.

The kernel capitalises the first kind and not the second. That is the only signal there is, and it is a reliable one across every step the loop has seen. Two exceptions of the same shape were a rule nobody had written down, so it is written down now and the special cases fold into it.

Verified live while a turn ran:

```
  Working
  Working
  Thinking · 5s
  Thinking · 5s
```

## The rest of the flow is whole

The same recording confirms the photo path end to end: `+` → `Photo Library` → pick → the chip appears instantly with its remove control → type → send → the photo drawn in the outgoing bubble above the words → the reply. Nothing clipped, overlapping or stuck, and the sentence the model gave is in the kernel's own record, so the photo described is the photo that was picked.

One thing checked and deliberately left: the three-second "Loading…" before the system picker is `PHPicker` loading the library, not ours — no such string exists in the app.

Findings M-421, M-422, M-423.


unarbos:
Merged into main as 9c00a389 (merge commit); rust fast-forwarded to the same SHA. TestFlight number will be reported only if this SHA's ios-testflight run prints `uploaded`.

unarbos:
The ios-testflight run for 9c00a389 (run 35388383077) printed `uploaded 0.2.0 (2117) 9c00a389`. TestFlight is now 2117.

PR #694, opened 2026-09-18
loop: give the mirror's refusal a way through, by name

unarbos:
Harness only.

`mirror-docs.sh` refuses to copy a store file that is shorter than the Mac's copy, because that shape is usually a truncation rather than an edit. That is the right default and it has caught real faults.

Twice today a **deliberate** shortening hit it — the build number's line rewritten more briefly — and both times I went around the refusal with a bare `scp`. That is strictly worse than the thing the refusal protects against, because `scp` skips every other check the mirror makes.

A refusal with no way past it is a refusal people go around. So there is a way through now, and it costs naming the file:

```
mirror: REFUSED mobile-mac-host-and-testflight.md — store copy 6456 b is shorter than the Mac's 6499 b.
mirror: that is the shape of the fault, not an edit. Read the store copy before deciding.
mirror: if you have read it and the shortening is deliberate, say so by name:
mirror:   SHORTER_IS_DELIBERATE=mobile-mac-host-and-testflight.md deploy/mobile/mirror-docs.sh mobile-mac-host-and-testflight.md
```

and with it:

```
mirror: mobile-mac-host-and-testflight.md is shorter and you said so — copying 6499 b → 6456 b
```

One file at a time, never a blanket flag, because the fault it guards against arrives one file at a time. Proven both ways by lengthening the Mac's copy on purpose and watching the refusal appear and then stand aside.

Findings M-424, M-425, M-426. M-425 records an audit that came up clean: after the "Working Thinking" fix, every file with several literal string special-cases was read, and the rest are deliberate formats — a `key: value` config parser and a markdown renderer — not exceptions standing in for a rule.


unarbos:
Merged into main as f7da9f06 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is f7da9f06.

PR #695, opened 2026-09-18
loop: the feedback poll uses the checkout it lives in

unarbos:
Harness only.

Two builds went out since the last poll, so this cycle polled TestFlight. **Nothing new** — 19 screenshots, 0 crashes. The two lines of shell error it printed above that report are what the cycle turned out to be about.

## It ran its poller from a second clone

`poll-feedback.sh` began with `cd ~/arbos-tools` — a second checkout in `$HOME`. That is the shape that cost this loop forty cycles of tooling, twice: the home copy usually works and silently lags.

On this machine the directory does not exist, so the `cd` failed, the `&&` chain stopped, and the run carried on in whatever directory it was started from. It worked only because that happened to be a checkout too.

It now resolves the repository it lives in. It also sourced `~/.op-env` without testing for it, which is where one of the error lines came from; it says where the credentials came from instead:

```
creds: the environment (no ~/.op-env on this machine)
```

## check-tools had called this tree clean all day

It matched `~/tool.sh` and not `cd ~/some-checkout`, which is the same reach wearing a coat.

Widening it went wrong twice, in opposite directions:

- `~/arbos` is the Mac's own checkout and must be exempt — but an exemption with **no boundary** also exempted `arbos-tools`, the single thing the widening was for, and the report went clean with my probe still sitting in the directory;
- the boundary then had to be "not a name character" rather than "a slash", because `cd ~/arbos &&` ends in a space.

Proven by planting a probe and watching it caught, then wrongly cleared, then caught again:

```
--- the real tree:              none — every tool resolves inside the repository
--- with a second clone planted: zz-probe.sh:1:cd ~/arbos-tools && echo hi
```

Findings M-427, M-428, M-429.


unarbos:
Merged into main as a29da225 (merge commit); rust fast-forwarded to the same SHA. The kernel red on the first run was subs_in_flight_e2e — unreachable from deploy/mobile — and passed on rerun. Mac worker: new main is a29da225.

PR #668, opened 2026-09-18
loop: reading the projects list needs the list as much as tapping it does

unarbos:
Harness only.

## "Inconclusive" every sweep for a week

`list-composer` has ended every sweep with `could not read both numbers — inconclusive`. Run on its own it passes all four steps:

```
--- 1. at rest ---         composer: Message pod…, and 'pod' is a row on screen
--- 2. with a filter ---   composer: Message subnet120…, it names one of them
--- 3. matching nothing -- composer: Plan, ask, build…, it offers no project, as it should
--- 4. above the keyboard  VERDICT: the composer sits 88 pt above the keyboard
```

Run second in the sweep it reads a **chat**: no rows, no composer, no keyboard. It needs the projects list and never asks for it — and [#662](https://github.com/unarbos/arbos/pull/662)'s check did not look at it, because it never *taps* a row. It counts rows and reads the composer's placeholder.

What kept it alive is the word. "Inconclusive" sounds like the world being unclear; here it meant the instrument was pointed at the wrong screen.

## The rule has now been wrong in both directions

- [#651](https://github.com/unarbos/arbos/pull/651) matched a string and missed four scenarios.
- [#662](https://github.com/unarbos/arbos/pull/662) matched a tap and missed this one.
- Widening it to "reads the list" flagged four files that were already correct, by counting helper *definitions* as uses.

It reads the main flow only now. And a scenario whose subject *is* the landing — `cold-start-and-history`, whose first step measures which screen a cold start opens on — declares itself exempt in a line of its own, rather than being bent to a rule it cannot satisfy.

Two more were quietly missing the step: `refusal-and-transport` kept its own inline copy of the idiom, `settings-and-a-bad-token` had none.

## The sweep, after

```
list-composer      the composer sits 88 pt above the keyboard
...
0 of 11 printed nothing at all.
0 of 11 printed output but reached no verdict — read those first.
```

Eleven scenarios, eleven conclusions. The one decline left is `pill-count-vs-sheet`'s `cannot say`, which is a scenario refusing a comparison it cannot make — by design.

Findings M-394, M-395, M-396.

[...]

PR #697, opened 2026-09-18
loop: run the edits I had not run — and three faults that came out of it

unarbos:
Harness only.

Between [#651](https://github.com/unarbos/arbos/pull/651) and [#668](https://github.com/unarbos/arbos/pull/668) the reach-the-list step went into about ten scenarios. Four were never exercised afterwards, and the sweep does not cover those four. This ran them.

Three pass — `sleeping-machine`, `photo-reaches-the-model` ("the model named the picture — the photo DID get through"), `worker-while-it-works`. **One was broken**, and fixing it uncovered two faults in the shared helper.

## `several-workers` taps its project twice

The first tap got the step at [#651](https://github.com/unarbos/arbos/pull/651); the one after "back and reopen" did not, and the run died there with `nothing matching 'phone' on screen`. One Back from the sheet lands on the chat, not the list.

The scenario also ended with `stills in …` and no verdict, leaving its own claim — that the workers survive leaving and coming back — for a reader to infer from two numbers. It says so now.

## `reach_the_list` reported success with a sheet over the list

It tested for the **absence of a Back button**. The workers sheet is a modal with no Back, so "no Back" read as "already on the list", and the caller tapped a row underneath it one line after a successful reach.

It succeeds on *seeing* the list now — the Projects header or a project row — and swipes down whatever is over it when there is no Back to press. This is the third time this one function has had the blind-success fault, and each time the same mistake: testing for the absence of the wrong thing.

## And it gave up one look too early

The loop checks, acts, sleeps — so the last action was never verified. It printed `cannot see the projects list after four tries` while the list was on screen a second later, because the fourth swipe had worked.

With all three fixed:

```
VERDICT: 4 of 4 named in the sheet, and the pill reads Agents 47 after
         going back and reopening — the workers are kept
```

## The check now states its boundary

`check-tools` reads the **first** use in a file, so a scenario returning to the list halfway through can fail exactly the way it guards against with the check silent — which is what happened here. That limit is printed with every pass.

Findings M-430 through M-433.


unarbos:
Merged into main as b82e2194 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is b82e2194.

PR #699, opened 2026-09-18
loop: the list's sections, which its coverage row has always named

unarbos:
Harness only.

The coverage row is "projects list — faces, rows, **sections**". `list-rows.sh` — which I wrote at cycle 112 and used to mark that row current — measures the rows and says nothing about the sections. Half the row's name had never been checked.

`list-sections.sh` checks the three claims `ProjectsView` makes:

- a header is drawn only when it has rows under it, because "an empty `Read` header over nothing reads as a section somebody collapsed, not as nothing matched";
- tapping a header folds its rows away and tapping again brings them back;
- a project with a live link belongs to `Working`, one without to `Read`.

The first two hold:

```
headers on screen:  Working absent, Read at y=175
rows on screen:     8
  under Read:       8

folding 'Read':
  rows after folding:   0   (was 8)
  rows after unfolding: 8
```

`Working` being absent is the behaviour — nothing is working — not a fault.

## The new check tested two of its own three claims

Membership needs both sections on screen at once, and every project rested in `Read`, so no row could be *seen* to be in the right one. The first draft would have printed a clean verdict having never looked. It says so instead:

```
Membership: not tested this run: only one section was on screen, so no row
            could be seen to be in the right one
```

Writing the claims into the header is what made that visible: the prose and the code disagreed, inside a file written to fix exactly that disagreement one level up.

Membership is read by position rather than by asking the app — a row between the `Working` header and the `Read` header is in `Working` — because that is what a person sees, and it cannot agree with the app by construction.

Findings M-436, M-437.


unarbos:
Merged into main as fc3f59a4 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is fc3f59a4.

PR #700, opened 2026-09-18
loop: the faces, which only pixels can see — and a rule I checked against the wrong key

unarbos:
Harness only. This finishes the coverage row "projects list — **faces**, rows, sections": rows were measured at cycle 112, sections at [#699](https://github.com/unarbos/arbos/pull/699), faces never.

The accessibility tree cannot see a face — a row is `const, Idle` whatever colour its folder is — so this reads pixels, which is the one thing in the harness that has to. It takes each row's position from the tree rather than hunting for rows in the image, which is what `find_row.py` did before it opened the wrong project twice.

## The faces hold

```
  const                              orange  held
  demo                               teal    held
  phone                              purple  held
  ...
distinct faces: 6 across 8 rows — blue, green, orange, purple, red, teal
  purple is worn by 2: parity-proj--reply-with-hostname, phone
  teal is worn by 2: demo, Clawdbot

VERDICT: every face held across two cold launches (8 rows, 6 distinct colours)
```

Two projects sharing a colour is ordinary with an eight-colour palette, so it is reported rather than counted against anything.

## The part worth reading

The first draft computed the name-derived colour — FNV-1a modulo the palette — and printed:

```
faces matching the name-derived rule: 0 of 8
```

which reads as every project wearing a face somebody chose. **The check was wrong.** `ProjectIdentity.defaults` hashes the *target* — `hub:<machine>/<project>` — and the list shows a machine for almost no row, so that key cannot be built from what is on screen.

Zero of eight should have been the giveaway. With eight colours, chance alone gives about one match; zero is the shape of a comparison that is not comparing. I committed it before doing that arithmetic.

The rule is now named as **unchecked** rather than checked wrongly, and the file says why, so the next person does not reach for the bare project name either.

Findings M-438, M-439.


unarbos:
Merged into main as d96e1039 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is d96e1039.

PR #701, opened 2026-09-18
loop: the files half of the attachments row, and why it cannot be driven

unarbos:
Harness only.

The coverage row is "attachments (`+`), photos, **files**". Every measurement under it is about photos — the chip check, the photo-reaches-the-model check, the journey's P2. No scenario had ever opened the Files picker.

## Attaching a file cannot be driven here

That is the useful finding, not a shortfall in effort. The picker opens on an empty `Recents`, and `Browse` reads **"On My iPhone is Empty"**. The app does not declare `UIFileSharingEnabled`, so a file dropped into its own Documents — which this cycle tried — is nowhere the picker can look.

Making it visible means changing the app to suit the harness, which is the wrong way round. So the gap is written into the check's verdict rather than worked around:

```
VERDICT: the + menu offers Files, the picker opens and closes, and
         cancelling leaves the composer as it was.
         Attaching a file is NOT exercised: the picker has nothing to
         offer on this simulator … That half of the row stays unmeasured,
         and says so.
```

## What is now checked

- the `+` menu offers **Files** as well as Photo Library;
- the picker opens — detected by the app's own controls disappearing, since the picker is another process the tree cannot see into;
- **the way out is clean**: the composer returns, no chip is left behind, and the typed text is unchanged. That one has history — a picker left open in journey run 32 swallowed the photo line and the whole call step after it.

## A ledger slip found on the way

`call — barge-in` still read **77** in the coverage table, though cycle 124 measured it (192 ms and 335 ms) and wrote the finding. The table was under-reporting the loop's own coverage; it is corrected here.

Reading every row's last cycle in one list is worth doing occasionally for its own sake. That table is the only thing that says what has gone stale, and it is kept by hand.

Findings M-440, M-441.


unarbos:
Merged into main as 1a0b17aa (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 1a0b17aa.

PR #702, opened 2026-09-18
loop: the orb's colours, which its row names and its check never read

unarbos:
Harness only. Closes the gap named at the end of [#701](https://github.com/unarbos/arbos/pull/701).

The coverage row is "call — voice first, orb, **colours**". `orb-phases.sh` read the phase *labels* out of the accessibility value and never looked at a colour — the word appeared only in its header comment.

That matters because a caller in a quiet room has the colour and nothing else. An orb that changed its accessibility value while looking identical would be a real fault, and the sequence check cannot see it.

## Measured, once per phase as it is entered

| phase | on screen |
|---|---|
| listening | RGB (229, 229, 229) |
| thinking | RGB (178, 178, 178) |
| speaking | RGB (128, 167, 218) |
| connecting | RGB (178, 178, 178) |

The three a caller meets mid-call are distinct — bright, dim, and blue — not only in the tree.

## connecting and thinking are the same grey, and it does not matter

My first version reported that flatly. It would have sent the next reader after a colour nobody has to tell apart: `connecting` happens once, before the call is under way, so no one is ever choosing between the two by eye.

Two **mid-call** states sharing a face would be the real fault, so the check says which kind it found:

```
  LOOK THE SAME: connecting and thinking
  each pair involves a state that happens once, before the call is under way,
  so nobody is choosing between them while listening. Worth knowing, not a fault.
```

That distinction is the point of the change. Finding two identical numbers is easy; saying whether they matter is what keeps a finding from becoming somebody's wasted afternoon.

Findings M-442, M-443.


unarbos:
Merged into main as 11e991cc (merge commit); rust fast-forwarded to the same SHA. The kernel red on the first run was browser_takeover_e2e — unreachable from deploy/mobile — and passed on rerun. Mac worker: new main is 11e991cc.

PR #698, opened 2026-09-18
loop: a committed check that a chat opens at its end

unarbos:
Harness only. No app change — the thing it was asked about does not reproduce.

## Measured, as asked

The standing evidence note says opening a chat with workers still clips by about 93 pt, and asks for a new measurement before a third guess at it. On `main` `b82e2194`:

| chat | unused scroll on opening | last line vs the pill |
|---|---|---|
| `phone` | **1 pt** | 87 pt above it |
| `demo` | **0 pt** | 109 pt above it |

A still confirms it — the last line reads whole, well clear of the pill. [#638](https://github.com/unarbos/arbos/pull/638)'s second scroll and [#642](https://github.com/unarbos/arbos/pull/642)'s bottom content margin hold together.

**The 93 pt was itself a bad number.** It came from cycle 110, where I took a single swipe for the end of a transcript. The real shortfall then was **501 pt**, which is what #638 actually fixed.

## The measurement is a file now

A hand measurement that settles a design question should not stay a memory. `chat-opens-at-the-end.sh` separates the two claims that were being conflated:

- **landing** — opening scrolls to the end, with no unused scroll left underneath;
- **clearance** — the last line is not under the workers pill, which floats over the transcript.

A chat can land at its end and still be clipped. Conflating those is exactly how [#633](https://github.com/unarbos/arbos/pull/633) shipped a fix aimed at the wrong mechanism.

It pages four times rather than once — one swipe is what produced the wrong 93 in the first place — and refuses to speak when it cannot read both positions.

```
== phone ==
  unused scroll:         1 pt
  landing:   lands at the end
  clearance: the last line sits 87 pt above the pill

VERDICT: every chat opened at its end with its last line clear of the pill (2 checked)
```

Findings M-434, M-435.

[...]

PR #704, opened 2026-09-18
loop: measure the streaming this file has photographed for eighty cycles

unarbos:
Harness only.

The coverage row is "project chat — send, prompt card, **streaming**, Worked line". `the-core-chat-path.sh` saves a still called `02-streaming` and measures the card, the first word and the Worked line. A reply that arrives in one lump passes all three exactly as well as one that grows word by word — only the caller sees the difference, and they see it for the whole length of the answer.

## It took three attempts, and the numbers caught the first two

1. **Sampled after the first word.** By then a three-sentence answer is already whole. It reported "only ever caught at one length" — honestly, which is how the timing flaw showed.
2. **Watched from the send, but took the last line matching a word the reply might start with.** That matched the *previous turn's* reply too, and reported **243 characters then 123**. A length going down is not a reply growing; that nonsense is the only reason the flaw was visible at all.
3. **Measures the transcript's total text**, which only grows while a turn runs and needs no guess about the reply's first word.

```
  the reply starts:        6.9s
  the transcript grows in: 2 step(s)   299 → 549 characters on screen
```

A run that catches only one length says it cannot tell streaming from a whole reply, rather than implying either — a short answer can finish between two samples, and one `ui dump` costs about 0.4 s.

## Also here

The coverage row's standing caveat — "opening a chat with workers stops 93 pt short" — is closed. Cycle 134 measured 0–1 pt of unused scroll with the last line 87 pt clear of the pill, and [#698](https://github.com/unarbos/arbos/pull/698) made that a committed check.

Findings M-444, M-445, M-446.


unarbos:
Merged into main as 55c82877 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 55c82877.

PR #705, opened 2026-09-18
loop: read the settings sheet's numbers instead of printing them

unarbos:
Harness only.

M-306 measured the sheet at cycle 87: five sections, three `Token saved` fields, a build line at the foot. Every run since has **printed** those numbers into a log that nothing compares. A section quietly lost would have appeared in the output of a passing run — the same shape as the voice-note check printing the words it never read.

They are checked now, against cycle 87's own counts:

```
  sections: Settings; Voice; Arbos kernel; Mesh hub; Notifications;
  section count: 5   (M-306 counted 5 at cycle 87)
  saved-token fields say: 3   (M-306 counted 3)
  build line: Arbos, 0.2.0 (1)
```

All three hold. A missing build line is a **fault** rather than a printed `MISSING`, because it is the only place the phone says which build it is running.

## The build line cannot answer the question this loop keeps being asked

On the simulator it reads `(1)` — the locally built number — under the words "The TestFlight build on this phone". That is true on a real install and a trap here, so the check says so when it sees it:

```
  (that is the local build: on the simulator this line cannot say
   which TestFlight build anyone has — only a real install can)
```

Most of today's traffic has been about which TestFlight number is current. This line looks like the place to settle it and is not, which is worth a printed sentence rather than a reader's assumption.

Findings M-447, M-448.


unarbos:
Merged into main as 2a95a48e (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 2a95a48e.

PR #706, opened 2026-09-19
loop: put the new checks in the sweep — and the first run found a regression

unarbos:
Harness only.

Eight checks have been written since the sweep's list was set, and none of them were in it: `list-rows`, `list-sections`, `list-faces`, `worker-chat-shape`, `chat-opens-at-the-end`, `attach-a-file`, `sleeping-machine`, `refusal-and-transport`. They ran only when somebody remembered, which is exactly how four scenarios went seventeen cycles without a run — three passed and one had been broken the whole time.

The sweep is nineteen now, and it came back **19 scenarios, 19 conclusions**, nothing silent and nothing verdict-less.

## What stays out, and why

Written into the file so nobody has to work out whether an omission was deliberate:

- `barge-in`, `orb-phases`, `one-breath-one-answer` — each drives a real call with injected audio, three to five minutes apiece, and needs the live gateway answering;
- `several-workers`, `worker-while-it-works` — watch a worker for two minutes by design;
- `what-a-returning-user-sees` — sleeps 120 s twice on purpose, because it is about coming back later;
- `type-send-measured` — needs a composer already open, and says so;
- `photo-reaches-the-model` — drives the system picker by coordinate, the one thing here a different simulator would break silently;
- `connect-times-by-engine` — reads the whole log history rather than the app.

## The first run found a regression

`list-rows` failed:

```
  PUNCTUATION:     phone: '·'
```

That is the fault [#642](https://github.com/unarbos/arbos/pull/642) fixed, and cycle 128 read the row cleanly. It now reads `phone, Idle,  · , home, 39m` again with `.accessibilityHidden(true)` still in the source, untouched. The app-side fix is batched in [#707](https://github.com/unarbos/arbos/pull/707).

That is the case for this change making itself: the check existed, it worked, and nothing was running it.

Findings M-449, M-450.


unarbos:
Merged into main (merge commit 4b76e18d); rust fast-forwarded to main 75609403. Mac worker: new main is 75609403.

PR #707, opened 2026-09-19
mobile: the row's status line is one thing said once

unarbos:
The seventh `ios/` batch.

[#642](https://github.com/unarbos/arbos/pull/642) hid the projects list row's ` · ` separator from VoiceOver, and cycle 128 read the row cleanly: `phone, Idle, home, 48m`.

It is back. Cycle 141's sweep read:

```
phone, Idle,  · , home, 39m
```

with `.accessibilityHidden(true)` still sitting in `ProjectsView`, untouched.

This is the same failure as the worker line's glyph at cycle 117, where the same modifier on an `Image` inside an `HStack` did not remove it from the tree. A modifier on one child of a composed `Button` label takes once and then stops holding.

The form that worked there works here: make the status line a **single element** and write its label out.

```
.accessibilityElement(children: .ignore)
.accessibilityLabel(([stateWord] + details).joined(separator: ", "))
```

The row now reads `phone, Idle, home, 7m` — a state and a machine, no punctuation announced — and `list-rows.sh` passes.

Twice now a hidden-modifier fix has passed once and quietly come undone. The one-element form has not, in either place it has been used.

Finding M-450.

<div><a href="https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7c66cfa8-381e-5700-9d78-3129f338a4fa&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #712, opened 2026-09-19
loop: say how much was examined, not only how much was wrong

unarbos:
Harness only.

## Where the undone fixes live

Two accessibility fixes inside the projects-list row came undone with the code untouched ([#642](https://github.com/unarbos/arbos/pull/642) clean at cycle 128 and dotted at 141; [#707](https://github.com/unarbos/arbos/pull/707) clean at 141b and dotted at 142). Two of the same shape elsewhere — the chat's worker-line glyph and the away card's bullet — have held since cycles 117 and 109.

The difference is the `Button`. A button builds its label by walking its children, and that walk varies between runs; the chat's rows are not composed button labels. That rule is now in `check-names.sh`'s header: **when it flags something under a button, say it on the button.**

I re-checked the two that held across three runs, with two worker lines and an away card **provably on screen** — clearing a chat that lacks the thing being cleared is exactly how cycle 117 fooled me.

## The count found something within the hour

`check-names.sh` now prints how many controls it examined, because "every control here is named" over three elements is a different sentence from the same words over thirty. The settings sheet read:

```
== the settings sheet ==
  every control here is named   (1 examined)
```

**One.** That screen is six text fields, five headings and a button; the detector looked only at buttons, images and pop-ups. A text field labelled by a symbol is exactly the fault this file exists for.

Widened to include fields and headings, the same screen reads **(12 examined)** — and the list 18, the workers sheet 13, the chat 7, the call 3.

The count was added for one reason and caught something else an hour later. That is the argument for printing what an instrument looked at, not only what it concluded.

Findings M-453, M-454.


unarbos:
Merged into main as e5e66f71 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is e5e66f71.

PR #713, opened 2026-09-19
loop: two verdicts that claimed what they only printed

unarbos:
Harness only. Continuing the audit of verdicts that can fire on a measurement that never happened.

## "and the composer cleared"

`the-core-chat-path.sh` ends its verdict with that phrase, and only ever **printed** the composer's contents. A send that left the typed line sitting in the box would have produced the same sentence — and that is a failure this loop has actually seen: a line typed and not sent, or sent and not cleared, is where M-162 started.

It is judged now: cleared means the composer holds a placeholder rather than the words that went out.

**And the judgement tests itself before it runs.** Feed it a placeholder and it must say cleared; feed it the line just sent and it must say not cleared. If it cannot tell those apart it declines to judge rather than guessing:

```
  the cleared-composer test cannot tell the two apart — not judging it
```

A branch that has never fired is a branch nobody has read.

## "never s to a list of 7 rows"

`cold-start-and-history.sh` printed `${FIRST:-never}s to ${LANDED:-nothing}`. On a timeout that reads as two claims that cannot both hold: nothing was timed, and the landing was read afterwards. A timeout now says so and names neither.

Neither fault would have appeared in a passing run. Both files print their evidence; the question each time is whether anything reads it.

Findings M-455, M-456.


unarbos:
Merged into main as 8b5d8654 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 8b5d8654.

PR #717, opened 2026-09-19
loop: what is behind the chat's ··· menu

unarbos:
Harness only.

Three items have sat behind that button since it was built — `Call <project>`, `Reconnect`, `Settings` — and no scenario had ever opened it. The button's own name was checked at cycle 108; what is under it was not.

```
== what the menu offers ==
  Call phone;Reconnect;Settings;
  the call item names its project: 'Call phone'

== Settings, from the same menu ==
  the settings sheet opened
```

The call item naming its project is the claim worth holding: "Call" and "Call phone" are different sentences when somebody with several projects open is about to talk to one of them.

**Reconnect is judged by what the chat does**, not by the tap being accepted — a reconnect that changes nothing and a menu item that does nothing are indistinguishable from the tap. A missed transition is reported as saying nothing either way, because a reconnect over a live link can finish inside one sample; a chat that never comes back is the fault.

It is in the sweep, so it runs without anyone remembering it.

## Two bugs of mine, loud in the log and silent in the verdict

- `"Opening $ROW…"` — bash reads `$ROW…` as a variable name, so the reconnect loop died with `ROW…: unbound variable` twenty times over **while the verdict printed as though it had watched**;
- `ui dump | grep -q` closes the pipe on its first match, and the reader died of `BrokenPipeError` mid-scenario.

Both fixed — the variable braced, the dump captured before it is searched. Errors in the output with confidence in the conclusion is the exact combination this loop keeps removing, and I managed to write a fresh instance of it inside the file meant to close a gap.

Findings M-459, M-460.


unarbos:
Merged into main as aab75652 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is aab75652.

PR #718, opened 2026-09-19
loop: the call's menu, and four ways I accused the app of my own bugs

unarbos:
Harness only.

The call's `≡` menu had never been opened by anything, like the chat's in [#717](https://github.com/unarbos/arbos/pull/717). It holds `Back to chat`, a route toggle and `Hang up`.

```
== the menu, mid-call ==
  offers: Back to chat;Use headset;Hang up;
== back to the chat ==
  landed in a chat, composer and all
```

Three runs, clean.

## The toggle cannot be judged here, and that is the finding

Its label names the route you are *not* on, which is one `!` away from telling every caller to switch to what they already use — nothing on screen would look wrong.

But `toggleSpeaker` flips a preference and then reads back what CoreAudio actually did, and a simulator has one output with nothing to switch to. The label is **right** not to move. That half belongs to the "AirPods / speaker route" row, which needs the phone, and the check says so rather than pretending.

## Four versions, three of them blaming the app

1. the label did not flip, so — fault. It cannot flip here.
2. judged against the route instead, but the route was unreadable, so the chain fell through to `FAULT: the route changed to ''` — an accusation assembled from an empty string.
3. `offers: nothing` from a menu it had **just listed**: it looked once to decide the menu was up and again to read it, and the menu closed in between.
4. tapped a menu that had closed itself after the toggle, then called the missing chat a fault. One run in two.

Each printed a confident sentence about the app. The rule underneath is small: **unknown must not fall through to an accusation.** Missing evidence gets its own branch, and it comes first.

Findings M-461, M-462.


unarbos:
Merged into main as 1ec3de64 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 1ec3de64.

PR #719, opened 2026-09-19
loop: does Stop end the turn, or only the screen?

unarbos:
Harness only.

The journey's J5 has carried the note *"Stop not exercised by this step (the control exists)"* since it was written, and nothing else touched it. That control exists because the kernel advertises it — its stall line reads **"Stop ends the turn"** — so a Stop that only quietened the phone would be the app repeating a promise it does not keep.

From the screen, those two are identical. This is judged on the kernel's record.

Against a `sleep 75`:

```
  the send disc became Stop: yes
  the square went away:      yes

== what the kernel did ==
  transcript right after Stop: 3448
  and thirty seconds later:    3448
     3446 user          stop 030429: run the bash command sleep 75 and nothing else, then reply done.
     3447 interrupted
     3448 turn_complete
  the turn is closed in the record: yes

VERDICT: Stop ends the turn — the square goes, the kernel's transcript
         stops growing, and the turn is closed in its record
```

The `interrupted` line is the kernel's own word, and the count not moving across thirty seconds is what rules out a turn that merely looks stopped.

The other branches say what they see rather than assuming: a square that goes while the transcript keeps growing is named as the app not keeping the kernel's promise; a transcript that stops without a `turn_complete` is reported as "stopped, or stalled and indistinguishable from here".

In the sweep, so it stays pressed.

Finding M-463.


unarbos:
Merged into main as 9821dc90 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 9821dc90.

PR #722, opened 2026-09-19
loop: the tool fold, and a character the rig silently drops

unarbos:
Harness only.

The tool fold came from Jacob at build 956 — *"I don't want to see these tool calls on the phone unless maybe I expand"* — and nothing had checked it since. `fold37.sh` is cycle-37 scratch: it reads `~/mobile-out/cycle-37` and takes its tools from `$HOME`, so it cannot run today.

It holds. Consecutive calls collapse to `2 tool calls, · 1 failed`, and opening adds two rows.

## The rig fact worth more than the check

The first run reported "no fold line appeared" — and the kernel's record showed my request had never arrived. **`idb ui text` types nothing at all for a line containing any non-ASCII character:**

```
  "plain ascii line"        -> non-ascii bytes: 0   … lands
  "with an em dash — here"  -> non-ascii bytes: 3   … composer still shows its placeholder
```

My line had an em dash. The scenario sent an empty box, got no answer, and blamed the fold.

`type_line` in `sim-lib.sh` now refuses a non-ASCII line before typing and reads the field back after — so M-162's read-back lesson and this trap are one helper instead of a habit every scenario has to remember.

## Three more of my own faults, each reading as a fault in the app

- **Counting rows across the toggle while the reply was still arriving.** The "1 row left behind" after closing was the model's answer landing. Counts are only comparable over a transcript that has stopped moving, so it waits for the Worked line.
- **Calling a one-call fold's empty expansion a fault.** The summary line *is* that call, and an opened fold shows labels rather than output (F14). With fewer than two calls it now says opening was not exercised.
- **A `case` pattern with a character range that refused a pure-ASCII line** — collation, not content. Counted bytes instead.

Findings M-464, M-465.


unarbos:
Merged into main as 6e7d1a01 (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 6e7d1a01.

PR #723, opened 2026-09-19
loop: the typing lesson lives in a helper, not in one file's comments

unarbos:
Harness only. Follows [#722](https://github.com/unarbos/arbos/pull/722), which lost a run because `idb ui text` types nothing for a line containing a non-ASCII character.

## Who else does it

Audited every `idb ui text` in the harness: **no literal non-ASCII anywhere**. But eight scenarios type a *variable*, so the protection has to be at runtime rather than at review.

## The lesson was already written down

`call-pulled-down.sh` has carried this since it was written:

> Plain ASCII only. `idb ui text` maps characters to key codes and throws "No keycode found" on anything outside the keyboard — an em dash in this very line cost a run.

It sat in that file and protected that file. A comment is read by whoever opens the file it is in; the next scenario is written by somebody who has not.

## And the scenario I wrote yesterday typed and sent blind

`stop-a-turn.sh` had **no read-back at all** — the fault M-162 named, which every other scenario had already fixed inside its own body. It was written the day before `type_line` existed and not revisited when it did.

Both now use `type_line`, which refuses what idb drops and reads the field back. Both still pass:

```
stop-a-turn:       3484 user → 3485 interrupted → 3486 turn_complete
                   transcript flat across thirty seconds
call-pulled-down:  pulled down it types, mutes, and closes back to the chat it came from
```

A lesson kept as prose protects one file. A lesson kept as a function protects the ones not written yet.

Findings M-466, M-467.


unarbos:
Merged into main as 24cdf35f (merge commit); rust fast-forwarded to the same SHA. Mac worker: new main is 24cdf35f.

PR #711, opened 2026-09-19
mobile: the row says its sentence on the button, not inside it

unarbos:
The eighth `ios/` batch, and the third attempt at one small thing.

Twice a fix applied **inside** the projects-list row read cleanly once and came undone a run later with the code untouched:

- [#642](https://github.com/unarbos/arbos/pull/642) hid the ` · ` separator — clean at cycle 128, dotted again at 141;
- [#707](https://github.com/unarbos/arbos/pull/707) made the status line one element with a label — clean at 141b, dotted again at 142.

A `Button` builds its label by walking what is under it, and that walk is what varies. So the sentence is stated **on the button**, from the row's own parts — name, state, machine, age:

```
phone, Idle, home, 13m
```

Nothing underneath can compose a different one. Three runs, eleven rows each, all clean.

## And the trait had to come back

`.accessibilityElement(children: .ignore)` takes the button's **traits** along with its children, so for a build the rows stopped announcing themselves as buttons. VoiceOver would no longer call a tappable row a button — and every harness check that finds rows by looking for Buttons found none, then reported "every row names a state" over an empty set. I used five such runs as proof the fix worked before noticing.

`.accessibilityAddTraits(.isButton)` puts it back. The check that was fooled is fixed in [#710](https://github.com/unarbos/arbos/pull/710).

Findings M-451, M-452.

<div><a href="https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7c66cfa8-381e-5700-9d78-3129f338a4fa&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #726, opened 2026-09-19
loop: a clean tree is not a clean binary

unarbos:

## What happened

The sweep ran all twenty-one scenarios. Every one reached a conclusion. One failed: `list-rows`, on `phone: '·'` — the separator [#711](https://github.com/unarbos/arbos/pull/711) removed, on a `main` that contains #711.

I measured rather than guessed: **6 of 6 dotted**. Then rebuilt from the *identical commit* by hand: **6 of 6 clean**.

The tree was right and the app was not.

## Why

This loop's own habit. Scp a modified source to the Mac, build, then `git reset --hard` it away. The restored file can land with an older timestamp than the object built from the scratch version, and the incremental build keeps the object. So `mac-cycle.sh` checked out the right commit, printed the right sha, and compiled an app from something else.

## The fix

When the reset discards anything, the built products go with it. One full compile in that case, nothing the rest of the time.

Verified end to end: `cleared 2 local change(s)` → rebuild → **6 of 6 clean**.

## It rewrites three findings

M-451 said two accessibility fixes inside the row "held once and came undone." M-452 was the third attempt after them. All were measured on builds taken straight after a scp-and-reset — and a stale binary produces exactly that story.

A probe label reached the tree intact (`PROBE-phone, Idle, home, 8m`), which is the evidence that the modifiers were working the whole time.

**The app fixes stand. The account of them was wrong.** Any measurement from cycles 141–150 taken right after a scp-and-reset is suspect until re-taken.

That is the most expensive fault this loop can have: not a wrong answer, but a wrong answer that looks like the app misbehaving, three times over.


unarbos:
Merged into main as 17792792 (merge commit) on the coordinator's word, CI green on 8b66d447; rust fast-forwarded to the same SHA. Mac worker: new main is 17792792.

PR #729, opened 2026-09-19
loop: withdraw #726's remedy, and ask whether the app on the device is this code

unarbos:

**This withdraws the fix I shipped yesterday in [#726](https://github.com/unarbos/arbos/pull/726). It was a remedy for a cause I could not reproduce.**

## What #726 claimed

Cycle 151 found a real symptom: a clean checkout of `main` produced an app that behaved like code from before a fix that checkout contained — six runs of six — while a rebuild from the identical commit read clean.

I blamed timestamps surviving a `git reset --hard`, and #726 deleted the built products whenever a reset discarded anything.

## Why it is withdrawn

I tested the claim directly.

- Put a visible marker in `ProjectsView.swift`, build → **12 rows carry it**
- `git reset --hard`, rebuild **with nothing deleted** → **0 rows carry it**

The reset is picked up correctly, every time I ask. Deleting `Build/Products` also recompiled nothing (`0` files), so what appeared to verify the remedy at cycle 151 was the manual full rebuild that happened just before it.

The symptom was real. Its cause is still unknown, and a remedy for a cause you cannot reproduce is a guess.

## What replaces it

Two questions that were never asked.

**Did the install work?** `xcrun simctl install`'s result was never checked. An install that fails leaves every later run measuring whatever was already on the device — which fits the symptom exactly.

**Is the app the code?** The installed binary is now dated against the newest `.swift` file. Whatever makes the device disagree with the tree leaves an app older than the code, and the run stops there.

Proven both ways:

- clean run → `the app on the device is newer than every source file`
- touch one source → `CAUGHT — app older than ProjectsView.swift`

## Two faults in my own test of that check

It first read as passing when it should have caught, because `find -newermt "@<epoch>"` does not parse on BSD find — it matches nothing, which looks exactly like a pass. A reference file carrying the app's date is compared instead.

And I was testing by scp-ing `mac-cycle.sh` onto the Mac and running it, which cannot work: the script's own `git reset --hard` restores the script from git partway through. The branch is pushed and checked out before it is tested.

## The findings this touches

M-451 and M-452 — "two accessibility fixes held once each and came undone" — are still best explained by the device disagreeing with the tree. A probe label reached the accessibility tree intact, which is the evidence the modifiers were working all along. **The app fixes stand.** What has changed is that I no longer claim to know why the device disagreed.


unarbos:
Merged into main as a91521a6 (merge commit) on the coordinator's word, CI green on 430733a4; rust fast-forwarded to the same SHA. Mac worker: new main is a91521a6.

PR #730, opened 2026-09-19
loop: the pill and the sheet cannot disagree, so ask something that can

unarbos:

## The problem

The sweep has carried one standing "cannot say" for many cycles: `pill-count-vs-sheet`, last reading *"the sheet's paging found 26 of the pill's 34."*

The check's own comment had already answered itself. The pill is `chat.workers.count`; the sheet is a `ForEach` over that same array. **Two drawings of one number.** They cannot disagree about the app, so every "cannot say" it printed was about the rig — failing to page a scrolling sheet, or failing to tell two workers apart when their goals truncate alike. The one disagreement it did file, at cycle 74, turned out to be exactly that.

I looked for an independent referee first. The kernel has no roster verb (`history`, `read`, `frames`, `hello`, `feedback`), and the app builds its worker list from those same frames — counting them would reproduce the app's logic rather than check it.

## What it asks now

Three workers the run starts itself, each goal led by a tag no earlier run used. The pill must rise by three and the sheet must show all three.

A row dropped in the drawing fails it. A rig that cannot reach the end of the sheet fails it too, and says which.

First run:

```
the pill says:  Agents 35
asking for three workers tagged p051856
  the pill now says: 38   (was 35)
the sheet lists: 30 rows (paging converged after 5 page(s))
rows carrying this run's tag p051856: 3 of 3
    p051856 one, Done
    p051856 three, Done
    p051856 two, Done
VERDICT: the sheet shows all three workers this run started.
```

## The totals gap, explained exactly

Pill 38, sheet 30, five pages, zero duplicates on any single screen. The difference is eight — and the project holds eight more agents than it has distinct goals. Twins on *different* pages, which a per-screen duplicate check cannot see by construction.

That reads the rig honestly, so the numbers stay. They are printed below the verdict as context, and are no longer asked to judge the app.


unarbos:
Merged into main as f44b88e2 (merge commit) on the coordinator's word, CI green on 8263bf30; rust fast-forwarded to the same SHA. Mac worker: new main is f44b88e2.

PR #737, opened 2026-09-19
loop: the journey names the app it tested, and refuses a target that cannot be one

unarbos:

The journey row was the oldest in the rotation and the last run was 21 hours back, so this cycle ran it.

## The run is clean

Fifteen steps pass, two scored by eye, four unverified with a stated reason each — J4's mid-flight half, J5's Stop not exercised by that step, J8 needing a kernel restart the phone cannot cause, and PUSH off until the key exists.

The kernel was asked on the attach socket at both ends and answered `c3247332dc4e` both times. All three phone-only steps pass, with the model naming the photograph: *Ice plant flowers, predominantly magenta.*

In passing: the list row read `phone, Idle, home, now` — the separator stays gone on a build the loop made itself.

## Two faults, both in the journey

**Every run record ever written names the kernel and not the app.** The record says `"branch": "main@unknown"`. `APP_BUILD` was read with a default at the single place it is used, and set nowhere.

It now carries the checkout's branch and sha — `main@959e5960`. That is only the app's build while the installed binary is not older than the sources, so the run asks that question (the same one `mac-cycle.sh` asks after it installs) and writes the doubt into the record when the answer is no.

**It accepted a cycle number as a target.** I started the run by handing it `157`. The journey takes `pod` or `<machine>/<project>`. It accepted the number, walked six steps, and failed with `J1 FAIL no 157 row on the list` — which reads like the app lost a project rather than like an argument in the wrong place.

It now refuses a target that cannot be one, the same guard `connect-times-by-engine.sh` already carries for the same reason.

Evidence in `media/mobile/cycle-157/` and the 19 stills under `media/mobile/journey/0919-062153/`.


unarbos:
Merged into main as c52d0253 (merge commit) on the coordinator's word, CI green on 439688ba; rust fast-forwarded to the same SHA. Mac worker: new main is c52d0253.

PR #736, opened 2026-09-19
edit: a CRLF file keeps its line endings through a hashline edit

unarbos:
## What

A destructive act on the record, on a machine we did not choose: a Windows checkout, or a repository with `eol=crlf` in `.gitattributes`.

**Control on `main` `02564618`**: one line edited by `LINE:HASH` in a five-line CRLF file. The file came back with **0 CRLF endings and 5 LF**. `lines()` drops the `\r`, and the join on `"\n"` wrote every line back LF — so the diff was the whole file: blame gone, the review a wall of line-ending churn, `changes` reporting a rewrite the model never meant.

## Change (`arbos-engine/src/tools/hashline.rs`)

`line_ending(text)`: `"\r\n"` when the text has more CRLF endings than bare LF, else `"\n"`. `edit` joins with it and uses it for the kept trailing newline. The `read` anchors were never affected (`lines()` on both sides), and the classic `old_string`/`new_string` edit is a substring replace, so it already kept the file's endings.

## Tests

`hashline::tests::a_crlf_file_keeps_its_line_endings_through_an_edit`: 5 CRLF lines, one replaced by two → 6 CRLF and no bare LF; a LF file stays LF; a stray CR in a LF file does not make the rule; a mostly-CRLF file is CRLF. The `hashline`, `editdiff` and `apply_patch` unit tests pass.

<div><a href="https://cursor.com/agents/bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcc57cf8-d8e5-575b-8c61-7a42eda2b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #740, opened 2026-09-19
read/grep/edit: a source that is not UTF-8 reads with a note, is searched, and is refused for edit with what to do

unarbos:
## What

A machine we did not choose: a codebase with Latin-1 or Windows-1252 sources — an old C, PHP or Java tree with `é` in a comment. **Probe on `main` `cc369869`** with a two-line C file whose comment holds one Latin-1 byte:

- `read` → `stream did not contain valid UTF-8`. The model cannot read the file at all.
- `grep` for a line that is in the file → **0 hits** (the walk skipped it in silence; the index path did the same).
- classic `edit` → the same UTF-8 error; hashline `edit` → **"file not found"**.

A misreport in each tool. Cursor reads such files.

## Change

- `fs::read_text` / `decode_text` (re-exported from `arbos_engine`): valid UTF-8 as is; otherwise `from_utf8_lossy` and the count of invalid sequences (`utf8_chunks`). A NUL in the first 8 KiB is a binary file, named as such with its size and what to read it with instead.
- `read`: one line above such a file — `[not valid UTF-8: N byte sequence(s) shown as �, likely Latin-1 or Windows-1252. edit will not write this file …; change it with bash, or convert it first: iconv -f latin1 -t utf-8]`. The lines follow with their anchors.
- both `grep` paths (the walk in `fs.rs`, the index in `kernel/grep.rs`) decode lossily; a binary is not searched.
- `edit`, `edit_all` and the hashline `edit` refuse a non-UTF-8 file with the path, the count and the two ways out. An edit would have re-encoded every byte it touched and the diff would not have been the change (#736's shape); the bytes stay exactly as they were. The hashline path checks existence first, so "file not found" is said only when true.

## Tests

`fs::not_utf8_tests`: the Latin-1 file reads with the note and its two lines; `grep` finds `int x` in it (1 hit) and does not search the binary; both edits are refused with `iconv` in the message and the file's bytes are unchanged; the binary is named `is a binary file (10 bytes)`; a clean file has no note. The `fs`, `hashline`, `apply_patch` unit tests and `grep_skips_git_internals_e2e`, `tool_parity_e2e` pass.


unarbos:
Merged into main as 380d149f (merge commit) on the coordinator's word, CI green on d148561c; rust fast-forwarded to the same SHA. Mac worker: new main is 380d149f.

PR #741, opened 2026-09-19
loop: say which coverage rows the harness actually reaches

unarbos:

## Why

The rotation rule takes the oldest coverage row first, and a row's age is the last cycle that *named* it. Cycle 156 caught the rule pointing at rows the sweep had run days earlier — `list composer` reading as cycle 81, `cold start` as 86, both through the cycle-151 sweep.

A row the sweep covers ages on paper while being tested every run, so the loop's own scheduler keeps sending cycles where they are not needed.

## What this adds

Every scenario declares the row it serves as a `# COVERS:` line. `coverage-map.sh` reads those against the table:

```
coverage rows:             33
rows some scenario claims: 24
rows the sweep reaches every run: 18
```

The nine rows nothing claims are the genuinely untested ones: AirPods/CallKit and the TestFlight build (both need Jacob's phone), the microphone path, the two journey rows, `recording` (its scenario is in unmerged [#732](https://github.com/unarbos/arbos/pull/732)), the whole-app style pair, typing into the composer, and the loop's own machine.

## It caught its own fault first

It read the sweep's list from `SCENARIOS=`, which is the argument array, rather than `DEFAULT=`, which is the default list. The sweep section printed **nothing at all**, under a verdict saying everything matched.

It now refuses to report when it has read no scenarios — the difference between an empty answer and a false one. Proven the other way too: one mistyped declaration turns the verdict and exits 1.

## It reads and does not write

Having it stamp the current cycle onto every row the sweep touches would be one line, and it would be wrong. The rows would age correctly while saying nothing true about whether anyone looked at them. The ledger's worth is that a cycle *named* a row.

[The projects list, ten rows](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Fcursor%2Fstores%2Fbc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983%2Fmedia%2Fmobile%2Fcycle-158%2F01-the-list.png)


unarbos:
Merged into main as 1234f620 (merge commit) on the coordinator's word, CI green on 0ded1927; rust fast-forwarded to the same SHA. Mac worker: new main is 1234f620.

PR #743, opened 2026-09-19
read: a page streams, a whole read has a cap, and a valid UTF-8 file is not copied twice

unarbos:
## What

A long-project shape: a 200 MB log, a fixture, a SQL export in the tree, and the model calls `read` on it for a look.

**Probe on `main` `8af86842`, release build**: `read` with `offset: 1, limit: 20` on a 200 MB log read the whole file into memory and hashed every line to show twenty — **0.47 s and ~400 MB for 1,960 bytes of answer**. A whole read took 1.3 s and built a **237 MB body** for the transcript to spill. At 2 GB that is a laptop swapping and a turn stalled.

## Change (`arbos-engine/src/tools/fs.rs`)

- `read_lines_from`: `offset`/`limit` stream the file through a `BufReader`, skipping what comes before the offset and stopping at the last wanted line. Each line is decoded on its own (lossily when not UTF-8, counted on the page — #740's note appears only when the page has such a line). A line over 4 MB is cut and said. The line cut follows `str::lines` exactly (`\n` or `\r\n`; a bare trailing `\r` stays), so a page's `LINE:HASH` anchors equal a whole read's.
- A whole read over `READ_MAX_BYTES` (64 MB) is refused with the size and the ways in: `offset`/`limit`, `grep`, or bash (`head`, `tail`, `sed -n 'A,Bp'`).
- `decode_text` takes the `Vec<u8>`; `String::from_utf8` keeps the buffer, so a valid UTF-8 file is no longer copied a second time (#740 had introduced that copy). The kernel's index grep passes the bytes by value.

**Same probe after**: the 20-line page in **0.12 ms**; the whole read refused in 9 µs.

## Tests

`fs::read_paged_tests`: a 50-line CRLF file's page (offset 5, limit 4) equals lines 5–8 of the whole read anchor for anchor, with no `\r`; a page past the end is empty, not an error; a Latin-1 byte is counted only on the page that holds it; a 65 MB sparse file is refused whole with `is 65 MB, over the 64 MB read limit` and its page is named a binary. All 147 engine unit tests, `tool_parity_e2e`, `grep_skips_git_internals_e2e`, `archive_children_e2e` and `worktree_store_e2e` pass.


unarbos:
Merged into main as 18f397cb (merge commit) on the coordinator's word, CI green on 19c00801; rust fast-forwarded to the same SHA. Mac worker: new main is 18f397cb.

PR #750, opened 2026-09-19
loop: the sweep of twenty-four, and keeping the scene when the separator fires

unarbos:
Harness only. No `ios/` change, so no TestFlight upload from this PR.

## Four checks were out of the sweep with no reason written down

That is the exact state the sweep's own header exists to prevent — it warns that a check outside the list only runs when somebody remembers it, and names the cycle where four went seventeen cycles unrun.

- `tool-fold` (cycle 149) and `call-menu` (147) were plain oversights — in now
- `call-text-in-chat` and `first-words-on-a-slow-connect` drive a real call and belong out; they are named rather than simply absent
- the one-offs named after cycles 37–39 are named too, as a record of how something was measured once rather than as checks

`typing-into-the-composer` runs **first**, because everything after it that sends anything goes through the helper it checks.

**Twenty-four ran. None silent, none without a verdict.** `pill-count-vs-sheet` gave a real answer instead of declining — the cycle-154 rewrite working in place.

## The separator, and five explanations that died

**6 of 6 dotted** during the sweep, on a build the freshness check certified as current. **6 of 6 clean** an hour later from the same commit. A probe label reached the tree intact in the clean runs, so the explicit label is in effect then.

Eliminated this cycle:

| explanation | how it died |
|---|---|
| stale binary | the app was certified newer than every source file |
| a second row call site | there is exactly one, used by both sections |
| `spoken` inserting the dot | it joins with `", "` and nothing else |
| the build step | `mac-cycle.sh` produced a clean row minutes after a dotted one, untouched tree |
| two sections open | a captured run had 2 headers and a Working row, label clean |

**So there is no fourth fix here.** Three fixes have each held once, and each was judged on a one-line flag with nothing around it. `list-rows.sh` now keeps the tree, a screenshot, the row and section counts, and the date of the installed binary when it fires. Proven by forcing the judge to report a fault:

```
  the screen at the moment it fired is in .../list-rows/fault-102730
  (section headers:  2, rows reading Working: 1)
```

## The second fault

`tool-fold` showed nothing more when opened — four rows closed, four open — on a turn with two tool calls, one failed. Filed rather than chased: a fold with two calls may have nothing further to show, and the check cannot yet tell that from one that refuses to open.

Its label also reads `2 tool calls, · 1 failed` — the same stray separator, in a second place.

[The projects list, clean](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Fcursor%2Fstores%2Fbc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983%2Fmedia%2Fmobile%2Fcycle-161%2F01-the-list-clean.png)


unarbos:
Merged into main as b623968d (merge commit) on the green head ad814562; rust fast-forwarded to the same SHA. Mac worker: new main is b623968d.

PR #753, opened 2026-09-19
checkpoint: add -A gets six tries when a file is rewritten under it, and a crashed git's index lock is cleared — the turn keeps its tree

unarbos:
## What

Live QA finding, draft `d037f76151` (scenario `rw-10-rewind-while-another-git-commits-in-the-same-repository`, kernel `d644bdf0`): **1 of 6 checkpoints had no tree** while another process rewrote a file in the repository — the harness's own `terminal-0.txt` — and `git add -A` failed with `short read while indexing terminal-0.txt`. Each such loss is a turn whose `rewind files:true` is refused for want of a tree.

**Reproduced on `main` `b623968d`**: a file truncated and refilled under git's mmap does not just fail the add — it **kills git (SIGBUS, exit 135, nothing on stderr; git 2.43)** and leaves the scratch index's `.lock` behind. With a writer that never paused, 16 of 20 tree saves were lost; with one that wrote for 120 ms and rested, 3–4 of 12.

## Change (`arbos-engine/src/tools/git.rs`, `work_tree`)

- `add -A` gets `ADD_TRIES` (6) attempts, 50 ms apart and growing (about a second in all): the file is done being written a moment later, and the tree is taken then.
- The scratch index's `.lock` is removed between tries and at the end, so a crashed git does not refuse the next attempt (`Unable to create '….lock': File exists`) and leaves no litter in `.arbos/`.
- A writer that never rests still beats it — then the error stands as before: said, not guessed; `rewind files:true` refuses rather than deletes.

## Tests

`git::churning_file_tests::a_file_rewritten_under_add_does_not_cost_the_turn_its_tree`: a burst writer (8 MB rewritten for 120 ms, 600 ms rest). The control loop with no retries loses 3–4 of 12 saves (counted and printed, not asserted); with retries **0 of 12**, and no `index-scratch*` litter left. Ran the `git::` set four times and the full engine suite (149 tests) once, all green.


unarbos:
Merged into main as bec7284b (merge commit) on the green head d2474fef; rust fast-forwarded to the same SHA. Mac worker: new main is bec7284b.

PR #757, opened 2026-09-19
loop: the call row holds, and a fix I took back

unarbos:
Harness only — **`ios/` ends this branch untouched**, so nothing for the upload cap.

## The row holds

Two scenarios cover `call — first word and transcription`. Neither had run since cycle 88, and both are out of the sweep because they drive a real call.

- **one breath**: 3 of 3 runs give one transcript, one spoken answer and one *kernel* answer, 900/900 frames
- **first words**: 6 of 6 kept every frame, at connects of 1226–1564 ms

## A check that was too hard on itself

With no connect over two seconds the slow-connect check said *"this run says nothing about the case in question"* — having just measured six runs that lost nothing. It reports what it kept and over what range now, and names only the slow case as untested.

Fixing it, I named my counter `RUNS` — which is the scenario's own argument for how many runs to do — and reset it to zero before the loop read it. A six-run check quietly did **two**, and numbered the second one `0`.

## One project, two faces — measured, unexplained, unfixed

The call screen draws `phone` with a red folder; the list draws the same project purple.

| | glyph |
|---|---|
| list | `(118, 93, 209)` |
| call screen | `(229, 83, 61)` |

In code the list keys a project's face on `KernelTarget.stored` — `hub:<machine>/<project>` — while the call screen's fallback keys on `<machine>/<project>`. A different string, so a different hash.

**I fixed that key and it changed nothing.** The glyph stayed at `(229, 83, 61)`. So the fallback is not the path taken: the chat's roster identity supplies that face, and the two screens read identity from different sources.

**The change is reverted.** The finding is measured and stands; the fix does not, because I cannot show it doing anything.

[The call screen: phone with a red folder glyph](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Fcursor%2Fstores%2Fbc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983%2Fmedia%2Fmobile%2Fcycle-165%2F01-the-call.png)

## And I read an empty grep as success

`mac-cycle.sh` refused a branch that exists only on the Mac — `cannot fetch fix165` — and my grep for `BUILD` printed nothing, which I took for a quiet success. I then measured a screenshot of the *previous* app and concluded the fix had failed.

The guard did its job; I did not read it. The second reading, with the branch's real name, built and installed properly — and the colour still did not move, which is how the finding above got its answer.


unarbos:
Merged into main as 668b841e (merge commit) on the coordinator's word, CI green on 0cf4b64d; rust fast-forwarded to the same SHA. Mac worker: new main is 668b841e.

PR #761, opened 2026-09-19
loop: a check dead for sixty-five cycles, and a helper that was never sourced

unarbos:
Harness only. `ios/` untouched, so nothing for the upload cap.

## The check had rotted

The work queue named `chat — the call's words read back`, last driven at cycle 102. It did not run — it stopped at `project row not on screen`.

It tapped `<project>, Idle`, which was the entire row label when it was written. The row has since gained its machine and an age — `phone, Idle, home, 5m` — so the string never matched again, and every run since has stopped on that line.

That is M-314's fault a second time: a script tapping words the app no longer says. It reaches the list and taps the project by name now.

Re-driven, the row holds:

```
  the kernel's words:  ...This project (poems, sorting algorithms, and...
  rows marked Spoken: 9; in this turn the answer appears 1 time(s)
  VERDICT: this turn has one row for its answer — the rule holds
```

## Fixing it nearly broke something else

I gave `list-rows.sh` a `reach_the_list` call and it had **no `sim-lib` line**. The shell would have said command not found, the helper would have returned 127, and the `|| exit 1` beside it would have ended the run before it measured anything.

Meanwhile `check-tools`, which looked only for the words, reported that every scenario reaches the list first.

It now lists any scenario calling a shared helper it has not sourced — matching *calls* rather than bare words, because `pt` appears in a comment about pt coordinates in two old files and had them reported as callers. Proven by removing the line again:

```
scenarios calling a shared helper without sourcing sim-lib:
  list-rows.sh calls: reach_the_list
```

## And `list-rows.sh` was hand-rolling the helper

One `Back` tap and carry on regardless, where `reach_the_list` tries four times and confirms it can see rows. The scenario that guards the list's rows was the one scenario not using the list helper.


unarbos:
Merged into main as e8613d01 (merge commit) on the coordinator's word, CI green on 520f2bf8; rust fast-forwarded to the same SHA. Mac worker: new main is e8613d01.

PR #770, opened 2026-09-19
loop: the names check said "the call" and meant three of its four controls

unarbos:
Harness only. `ios/` untouched, so nothing for the upload cap.

## Three integrity checks, all clean

The machine can run the whole harness. Every tool resolves inside the checkout, every scenario reaches the list first, and none calls a helper it has not sourced. No control on any screen reads as a symbol name.

## But the names check was flattering itself

Its own comment says the call *"rests entirely on the labels of its four controls"*. It examined three — `Call menu`, `Settings`, and the orb, which reads `Call` because tapping it starts one.

The fourth is `End call`, and it does not exist until a call is up. So the screen was examined in the one state where its end control is absent, and "every control here is named" was true of three quarters of it without saying so.

**A preview call on this simulator will not connect.** With no clip the screen says `No microphone input.` and the orb does nothing at all; with an injected clip it accepts the tap and stays where it is. `End call` is tapped by name in `call-text-in-chat.sh`, on a call that is actually running — so the label is covered, just not from here.

It examines both states now and names the one it cannot reach:

```
== the call, at rest ==
  every control here is named   (3 examined)
== the call, connected ==
  not reached from a preview call, so End call was not examined here.
  It is tapped by name in call-text-in-chat.sh, on a call that is up.
```

## My first version was worse than the problem

It counted the unreachable state as a missed screen, so every run ended `VERDICT: incomplete`. A check that can never be complete is a check people stop reading. Withdrawn inside the cycle. An orb missing altogether is still a miss.

## The `ios/` batch, for whoever decides the next upload

It holds exactly two commits:

- `c108eecb` — moved the project row's label onto the button
- `7cb69c3b` — put back the `.isButton` trait that move removed

Nothing else has touched `ios/` since the 2160 build. The first is the separator work, whose effect this loop has measured as intermittent across six eliminated explanations; the second is a plain repair and is not.

[The call screen at rest](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Fcursor%2Fstores%2Fbc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983%2Fmedia%2Fmobile%2Fcycle-172%2F01-the-call-at-rest.png)


unarbos:
Merged into main as 3bccf924 (merge commit) on the coordinator's word, CI green on c21faf4a; rust fast-forwarded to the same SHA. Mac worker: new main is 3bccf924.

PR #767, opened 2026-09-19
loop: the style pair is a check, not a ritual

unarbos:
Harness only. `ios/` untouched, so nothing for the upload cap.

## Both surfaces hold

| surface | Arbos | Cursor | apart |
|---|---|---|---|
| list row pitch | **8.7%** of screen height | 8.3% | 0.4 |
| chat left margin | **5.3%** of width | 4.9% | 0.4 |

The same numbers cycle 83 got for the list and cycle 152 for the chat — sixty-nine and nineteen cycles ago, on phones of different sizes. Ground is printed and never compared; dark-only is deliberate on both sides.

## Why the row keeps ageing

`style-pair.py` has done the measuring since cycle 83. What it never had was anything to hand it the pictures — every reading of this row has been a person taking two screenshots and typing two commands.

So the row aged eleven cycles between readings, and cycle 152 spent an entire cycle discovering the tool could not even run on the Mac.

## What this adds

`style-pair.sh` captures both surfaces and judges both numbers:

```
  list row pitch:   8.7% against 8.3%  — 0.4 apart
  chat left margin: 5.3% against 4.9%  — 0.4 apart

VERDICT: both surfaces sit within a point of Cursor's — the list's rhythm
         and the chat's left margin
```

A point is the threshold: twice measured under half a point, sixty-nine cycles apart, so a whole point sits comfortably outside the noise and far tighter than an eye would call a difference.

It refuses a verdict on a list too short to have a rhythm, and says plainly that a still of the wrong screen measures perfectly and means nothing.

**Proven it can fail:** tightened to a tenth of a point, the verdict turns to `2 surface(s) are more than a point apart`.

[The projects list](https://cursor.com/agents/bc-7c66cfa8-381e-5700-9d78-3129f338a4fa/artifacts?path=%2Fcursor%2Fstores%2Fbc-ec8c092a-3084-4e3e-9e34-7b2a1f8c6983%2Fmedia%2Fmobile%2Fcycle-171%2F01-the-list.png)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

[...]

PR #771, opened 2026-09-19
dev channel: ask again on a timer, because a CI completion is not a reliable wake-up

unarbos:
## A green that no event will wake

The plan step decides well, and it only decides when something wakes it. Today, twice in one hour, nothing did.

A commit gets two CI runs. Both times the **failing** one finished first, fired this workflow, and was correctly told there was nothing green to publish. Then the **green** sibling completed and fired no event at all.

Measured on #770's commit `3bccf924`, not inferred:

| CI run | conclusion | finished | dev-channel run |
| --- | --- | --- | --- |
| `35447685644` | failure | 14:08:30 | fired at 14:08:32 |
| `35447686732` | success | 14:15:25 | **none** |

The 14:08 run said, correctly:

```
No commit on main newer than build 2288 has a green CI run, within the last 40,
so there is nothing to publish.
```

And that was the last word on it. #767's green (build 2286) sat unpublished for 38 minutes earlier in the same hour and went out only because the next merge's CI happened to fire. #770's green (build 2292) sat until this was written.

Neither was a wrong decision. Both were a right decision that nobody asked for a second time.

## Why the existing caps do not cover it

Every cap in this file is a clock read **at a decision**. #596 put a clock on the pending commit; #604 moved it onto the green build so that no new merge could reset it. Both make the decision better. Neither helps when the decision does not happen at all, which is the shape here.

## The change

The workflow also runs on a ten-minute timer, and the plan job accepts a `schedule` event.

A timer run has no triggering commit, so `EVENT_SHA` is empty, nothing is treated as this run's own, and every candidate is asked of the API. That is the only behavioural difference, and it is the correct one — the "never wait on yourself" shortcut from #517 exists precisely because a run *has* an event, and a timer run does not. It takes the ordinary walk, not the `workflow_dispatch` publish-what-you-were-pointed-at path.

## Cost

A plan job every ten minutes that on almost every run says there is nothing to publish and stops: one `ubuntu-latest` job, a `git rev-list`, and a few API calls. It buys a channel whose worst case is ten minutes rather than "until somebody merges something".

## Tests

Two cases for the timer's shape, in `.github/dev-channel-plan-test.sh`:

- a run with no event of its own still finds the green
- a run with no event of its own still goes back past a red tip

All twenty-three cases pass. The `publishing` job in CI runs them.


unarbos:
Merged into main as 768df3b7 (merge commit) on the green head ea0d3e68; rust fast-forwarded to the same SHA. The schedule trigger is live from this merge. Mac worker: new main is 768df3b7.

PR #776, opened 2026-09-19
loop: a fault whose fix is unmerged is still reported every run

unarbos:
Harness only. `ios/` untouched, so nothing for the upload cap.

[#774](https://github.com/unarbos/arbos/pull/774) fixed three scenarios that were installing a day-old build. This re-ran the same sweep with the fix in, to see which verdicts had been resting on that stale app.

## One changed, and it is the one that mattered

`list-rows` now reads *"every row names a state, ages read as ages, and nothing speaks a separator"* — where the identical check in cycle 173's sweep reported the separator. Same scenario, same position in the run, one change between them.

#774 is confirmed by a second measurement rather than only by the hand-run that motivated it.

## The other fault did not change, and my explanation of it was wrong

Cycle 173's report said `tool-fold` "was measuring that same stale app". **It was not.** It is main's copy of the *check* — whose repair (cycle 162: four separate faults in the check, none of them in the app) has been sitting in open [#751](https://github.com/unarbos/arbos/pull/751) ever since.

Both measured on one build:

| copy of the check | what it says |
|---|---|
| main | `opening the fold showed nothing more (4 → 4)` |
| [#751](https://github.com/unarbos/arbos/pull/751) | `'6 tool calls' hides 6 line(s) and gives them back` |

The sweep has been printing a fault that was disproved thirteen cycles ago.

## Two harness fixes are stranded in open PRs

[#751](https://github.com/unarbos/arbos/pull/751) for the fold, and [#732](https://github.com/unarbos/arbos/pull/732) for `record-workers.sh`, which I have checked out by hand for every recording since cycle 155.

The sweep's header now says to check whether a fault's fix is already written and merely unmerged before chasing it. The sweep runs what is checked out, and that is `main` unless a cycle says otherwise.

Twenty-five scenarios, none silent, none without a verdict, one fault — and it is a stale check.

[...]

PR #780, opened 2026-09-19
dev channel: a run stays and asks again, rather than ending on a promise

unarbos:
## Three greens, none asked twice

Every ending in the plan step was "somebody will come back here and ask". Three times today nobody did.

A commit gets two CI runs. The failing one finishes first, fires this workflow, and the plan correctly publishes nothing because the green sibling has not landed. Then the green sibling completes and **fires no event at all**. Measured on #770's commit `3bccf924`:

| CI run | conclusion | finished | dev-channel run |
| --- | --- | --- | --- |
| `35447685644` | failure | 14:08:30 | fired at 14:08:32 |
| `35447686732` | success | 14:15:25 | **none** |

- **#767**'s green (build 2286) sat 38 minutes, released only when the next merge's CI happened to fire.
- **#770**'s green (build 2292) sat until #771 was written.
- **#776**'s green sat until this was.

Each decision was right. None was asked a second time.

## Why not just the timer

[#771](https://github.com/unarbos/arbos/pull/771) added a ten-minute `schedule` for exactly this case. **It has not fired once in the two hours since it landed**, and this repository has never had a scheduled run of any workflow — checked with `?event=schedule`, which returns nothing for the whole repository. Not a fork, not archived, workflow `active`, cron on the default branch. It may simply be GitHub delaying a new high-frequency cron on a busy repository, and it may start working. The channel should not be left leaning on it either way.

This change needs nothing from GitHub's scheduler and nothing from the next merge.

## The change

The run that would have ended on the promise stays and asks again.

- `WATCH` (600 s, matching `PATIENCE`) bounds how long it stays. Past `WATCH` the green-clock cap from #604 publishes anyway, so staying longer wins nothing.
- `BEAT` (60 s) paces the asking.
- A run only stays while something is **actually running that might go green**. A red `main` with nothing in flight ends at once, because no amount of waiting changes it. That is what `pending_seen` is for.

Cost: one idle `ubuntu-latest` job while it waits. The `dev-channel` concurrency group already serialises these and drops all but the newest waiting run, so they cannot pile up.

## Tests

Three cases added. Two fail against the logic they replace:

```
FAIL  a run stays, and sees the green that lands while it waits — expected 4, decided none
FAIL  a run with nothing green yet stays for the one still running — expected 4, decided none
```

The third pins the refusal: *a run with nothing running does not wait for what cannot come.*

Every case already in the file runs with `WATCH=0`, which is the old behaviour exactly, so they pin what staying must not change. All twenty-six pass.

One fix to the harness came out of writing these. The new `late` stand-in remembers that it has been asked once, so its second answer can differ from its first — and that memory was leaking between cases, so a case written to fail passed because the *previous* case had already asked. `check` now clears it. Worth knowing, because it is the same shape as the bug being fixed: a state that outlives the run it belonged to.


unarbos:
Merged into main as 73c22fa4 (merge commit) on the green head 28689899; rust fast-forwarded to the same SHA. Mac worker: new main is 73c22fa4.

